PR #66, opened 2023-04-20
feat: eds provider and theming

### LINKED ISSUE #47: Create design interface
This is going to need some investigation + trial and error:

While we are waiting for EDS 2 and its new design tokens system, It would be a great idea to create our own interface for defining values used in our component package.
We are going to need a system that can provide the following

- [ ] #65
- [ ] - Automatically adjust to screen size (by breakpoints as used in web maybe?)
- [ ] - Expose all its values to a user. Any user of the component library should be able to make its own components with EDS values provided (colors, paddings, roundings etc). These should also abide to the dark mode / light mode setting

For light / dark mode, color values should be structured in a typesafe matter.

This issue covers the investigation and creation of a POC that demonstrates an elegant and performant method of achieving this with as little dependencies and boilerplate as necessary.

### LINKED ISSUE #65: Support switching between light mode and dark mode

HaakonSvane:
- Add EDSProvider 
- Add theming engine and a master token for storing all theming values
- Add `EDSStyleSheet.create`, `useStyles` and `useToken` method and hooks
- Convert all current components to use theming engine
- Remove `translations`, `constants` and `@equinor/eds-tokens`

closes #65
closes #47


HaakonSvane:
I should perhaps add some images:

<p float="right" >
Dark3

Dark1
 </p>

<p float="right" >
Light2

Dark2
 </p>

<p float="right" >
Light3

Dark3
 </p>

hathibelagal-dev2 (approved):

PR #69, opened 2023-04-20
feat: button group and toggle button

### LINKED ISSUE #67: Add ButtonGroup

hathibelagal-dev2:
Closes #67


packages/components/src/components/Button/Button.tsx
```diff
@@ -23,29 +25,72 @@ export const Button = React.forwardRef<View, ButtonProps & ViewProps>(
...
+        style.push(rest.style);
 
         return (
-            <View ref={ref} style={[styles.colorContainer, rest.style]}>
-                <PressableHighlight
-                    onPress={onPress}
-                    style={styles.pressableContainer}
-                >
+            <View ref={ref} style={style}>
+                {isToggleButton ? <View style={styles.pressableContainer}>
```
HaakonSvane, line 51: I see that you are defining a `Pressable` wrapper in the ToggleButton and omitting the `onPress` callback here. Would it be possible to expose a callback in the context named `onToggle()` or something, instead?
This way, you don't need two different views being rendered, only a conditional callback invokation: `onPress={toggleData ? toggleData.onToggle : onPress}`

packages/components/src/components/ButtonGroup/ButtonGroup.tsx
```diff
@@ -0,0 +1,51 @@
...
+    length: 0,
+    index: -1
+} as ButtonGroupContextContents);
+
+export const ButtonGroup = (props: ButtonGroupProps & ViewProps) => {
+    return (
+        <View style={props.vertical ? styles.vertical : styles.horizontal}>
+            {
+                Children.map(props.children, (child: ReactNode, i: number) => {
+                    console.log(i);
```
HaakonSvane, line 25: console.log here

packages/components/src/components/ButtonGroup/index.ts
HaakonSvane, line 1: These could be put in the same folder as the button, since they are specializations of the button component.
Also, perhaps look into exporting the ButtonGroup and ToggleButton as `Button.Group` and `Button.Toggle`. This makes for better discoverability and is more in line with the EDS react interface.
Refer to the EDS implementation for information on how to do this, or look in my implementation of the Accordion in the new-obervation branch of om-inspection

HaakonSvane (changes requested):
All in all, awesome job. The results look amazing and very easy to use.I'm glad you figured out a solution to this using contexts and not cloning 👍 Please see my comments for any improvements / discussion points

apps/chronicles/package.json
```diff
@@ -18,8 +18,8 @@
     "@react-navigation/bottom-tabs": "^6.0.5",
     "@react-navigation/native": "^6.0.2",
     "@react-navigation/native-stack": "^6.1.0",
-    "@shopify/react-native-skia": "0.1.172",
-    "expo": "~48.0.10",
+    "@shopify/react-native-skia": "0.1.183",
```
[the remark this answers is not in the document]
hathibelagal-dev2: I don't get any error with doctor. Yeah, guess we just have to work with different versions for now.

packages/components/src/index.ts
```diff
@@ -6,8 +6,8 @@ export * from "./components/Navigation";
 export * from "./components/PressableHighlight";
 export * from "./components/EDSProvider";
 export * from "./components/Spacer";
-export * from "./components/ButtonGroup";
-export * from "./components/ToggleButton";
+export * from "./components/Button/ButtonGroup";
+export * from "./components/Button/ToggleButton";
```
HaakonSvane, line 10: I think you don't have to export these any more as they are now part of the `Button` export (?)

HaakonSvane (changes requested):
Absolutely amazing 🔥

PR #70, opened 2023-04-23
fix: support packages not initializing provider

### LINKED ISSUE #68: Support Skia without EDSProvider

HaakonSvane:
- Remove not needed throw

closes #68


HaakonSvane:
Turns out we were already defining default values for the context, so nothing besides removing a not needed throw was done ;)

hathibelagal-dev2 (approved):
Indeed.

PR #77, opened 2023-04-27
feat: add inputs

hathibelagal-dev2:


packages/components/src/components/TextField/TextField.tsx
HaakonSvane, line 1: After playing around with it a little bit, I noticed that the selectable area of the text field is a little small in comparison with the rest of the component. What I mean is that there are areas of the gray container that will not respond to touch. Maybe look into expanding the text area of this area, if that makes sense

HaakonSvane (changes requested):
Great job! This looks very good. Code pretty much is in tune with the rest of the library as well. Awesome. 
For the technical stuff, I have left some comments for you to read.
When it comes to the design, I just had a talk with @EmilieOlsen regarding some visual changes.

- The label above the text inputs should be smaller in size (0.75*base), regular font and tertiary coloring.
- The left padding of the label text should be in alignment with EDS and should not be affected by the adornments as they are now
- The adornments should have the same background color as the text input so that they look more integrated with the text field

packages/components/src/styling/masterToken.ts
```diff
@@ -25,16 +25,16 @@ export const masterToken: MasterToken = {
...
             medium: {
                 light: "#DCDCDC",
-                dark: "#132634",
+                dark: "#3E4F5C",
             },
         },
         container: {
             background: {
-                light: "#F7F7F7",
+                light: "#FFFFFF",
```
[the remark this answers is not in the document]
hathibelagal-dev2: This should probably be moved to a new issue. I find it hard to map https://www.figma.com/file/0TbIXrrObWj80Cf7KucKYFL0/Design-Tokens?type=design&node-id=2-3&t=TXomYnfWoY2Yq11I-0 to masterToken.

hathibelagal-dev2:
I think by having that extra padding, we're forcing users to not give a background to the left and right adornments. Here's eds-core-react's Input with a background 
image

packages/components/src/components/TextField/TextField.tsx
```diff
@@ -11,55 +11,73 @@ export type TextFieldProps = {
...
-            marginBottom: 4,
+            marginBottom: 8,
             marginLeft: -12,
         },
         outerContainer: {
             marginLeft: 12,
         },
         textInput: {
             color: theme.colors.text.primary,
+            fontFamily: theme.typography.basic.p.fontFamily,
```
HaakonSvane, line 83: You can do

```ts
textInput: {
    color: theme.colors.text.primary,
    ...theme.typography.basic.p
}
```

to destructure everything in the font token. Makes the style a little bit more dynamic

packages/components/src/components/TextField/TextField.tsx
```diff
@@ -11,55 +11,73 @@ export type TextFieldProps = {
...
+            fontSize: theme.typography.basic.p.fontSize,
         },
         placeholder: {
             color: theme.colors.text.tertiary,
+            fontFamily: theme.typography.basic.p.fontFamily,
```
HaakonSvane, line 88: Same as the other comment regarding destructuring

packages/components/src/styling/masterToken.ts
```diff
@@ -254,7 +254,7 @@ export const masterToken: MasterToken = {
             },
             label: {
                 fontFamily: "Equinor-Bold",
```
HaakonSvane, line 256: This should be regular and not bold

HaakonSvane (changes requested):

PR #94, opened 2023-05-20
fix: cell styling

HaakonSvane:
- rework cell styling
- fix padding missing for labels in `Input` component
- Add new splashscreen to Chronicles
- Add adjusted Components header image


HaakonSvane:
The changes made to the cells are based on  feedback from @EmilieOlsen

PR #96, opened 2023-05-23
feat: Icon button

### LINKED ISSUE #89: Add support for icons in buttons
We should add support for a `iconPosition` (type `leading | trailing`) and a `icon` (type `IconName`) prop to the `Button component`
From Figma:
Screenshot 2023-05-10 at 22 55 25

Now might also be a smart time to look at how we expose the button. With all the different variants supported, it might make more sense to rather expose the Button component as a familiy. For example, we could expose the following components:
- `Button`: Normal button. Supports outlined, contained or ghost mode. Has an iconPosition prop and a icon prop like this issue discusses
- `Button.Icon`: Pure icon button. Supports same props as button minus the text label.

This makes props easier to understand and does not clutter the prop space of the components.

HaakonSvane:
- Refactor button component family
- Add icon support for text-labeled buttons
- Add `Button.Icon` component

closes #89

### DIFF
--- a/.changeset/curly-olives-admire.md
+++ b/.changeset/curly-olives-admire.md
@@ -2,4 +2,4 @@
 "@equinor/mad-components": patch
 ---
 
-Rework Cell styling
+Reworked Cell styling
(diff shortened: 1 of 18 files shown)


hathibelagal-dev2 (approved):
Nice!

PR #119, opened 2023-06-27
feat: add left and right swipe items for `Cell`

### LINKED ISSUE #111: Support for swipe left/right in Cells/CellGroup
I would love to start using mad-components in PleaseMove, but in my current navigation cells I have an option to add "swipe right to get option to delete" behaviour. This is not currently possible with mad-components' cells
image

Is this something that is possible to add in mad components? Would it even make sense to add it?

HaakonSvane:
closes #111

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -29,7 +29,7 @@
         "react-dom": ">=16.8",
         "react-native": ">=0.68",
         "expo-font": ">=10.0.0",
-        "react-native-svg": "*",
+        "react-native-svg": ">=13.0",
         "react-native-reanimated": ">=2.0",
         "react-native-gesture-handler": ">=2.0"
     },
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -29,7 +29,9 @@
         "react-dom": ">=16.8",
         "react-native": ">=0.68",
         "expo-font": ">=10.0.0",
-        "react-native-svg": "*"
+        "react-native-svg": "*",
+        "react-native-reanimated": ">=2.0",
+        "react-native-gesture-handler": ">=2.0"
     },
     "dependencies": {
         "@floating-ui/react-native": "^0.10.0"
--- a/packages/components/src/components/Cell/Cell.tsx
+++ b/packages/components/src/components/Cell/Cell.tsx
@@ -32,6 +32,7 @@ export const Cell = React.forwardRef<View, React.PropsWithChildren<CellProps>>(
     ) => {
         const { isFirstCell, isLastCell } = useContext(CellGroupContext);
         const styles = useStyles(themeStyle, { isFirstCell, isLastCell });
+
         const CellContent = () => (
             <View {...rest} style={[styles.container, rest.style]} ref={ref}>
                 <PressableHighlight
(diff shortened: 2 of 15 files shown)


packages/components/package.json
```diff
@@ -29,7 +29,9 @@
         "react-dom": ">=16.8",
         "react-native": ">=0.68",
         "expo-font": ">=10.0.0",
-        "react-native-svg": "*"
+        "react-native-svg": "*",
```
tormodAase, line 32: Will this package work with all of the 13 different major versions of react-native-svg?

packages/components/src/components/Cell/Cell.tsx
```diff
@@ -48,20 +55,38 @@ export const Cell = React.forwardRef<View, React.PropsWithChildren<CellProps>>(
...
+                    renderLeftActions={() => leftSwipeGroup && leftSwipeGroup.map((swipeItem, index) => (
+                        <CellSwipeItem key={`testL_${index}`} {...swipeItem} />
+                    ))}
+                    renderRightActions={() => rightSwipeGroup && rightSwipeGroup.map((swipeItem, index) => (
+                        <CellSwipeItem key={`test_${index}`} {...swipeItem} />
+                    ))}>
+                    {CellContent()}
+                </Swipeable>
+                :
+                CellContent()
```
tormodAase, line 73: `CellContent()` should probably be changed to `<CellContent />`

packages/components/src/components/Cell/Cell.tsx
```diff
@@ -48,20 +55,38 @@ export const Cell = React.forwardRef<View, React.PropsWithChildren<CellProps>>(
...
+        return (
+            (leftSwipeGroup || rightSwipeGroup) ?
+                <Swipeable
+                    overshootFriction={8}
+                    containerStyle={{ backgroundColor: styles.container.backgroundColor }}
+                    renderLeftActions={() => leftSwipeGroup && leftSwipeGroup.map((swipeItem, index) => (
+                        <CellSwipeItem key={`testL_${index}`} {...swipeItem} />
+                    ))}
+                    renderRightActions={() => rightSwipeGroup && rightSwipeGroup.map((swipeItem, index) => (
+                        <CellSwipeItem key={`test_${index}`} {...swipeItem} />
```
tormodAase, line 68: should probably replace "test" with something else in the key names

packages/components/package.json
HaakonSvane: Technically, no. I should perhaps force the major version at least

packages/components/src/components/Cell/index.ts
```diff
@@ -15,5 +16,6 @@ export {
     Cell,
     CellProps,
     CellGroupProps,
-    NavigationCellProps
+    NavigationCellProps,
+    CellSwipeItemProps as CellSwipeGroup
```
tormodAase, line 20: I don't know how I feel about renaming `CellSwipeItemProps`.

tormodAase (changes requested):

packages/components/src/components/Cell/index.ts
tormodAase: If you import from index, it is called CellSwipeGroup, but if you import from types, it is called CellSwipeItemProps. But the types are identical. This can be confusing

packages/components/src/components/Cell/Cell.tsx
HaakonSvane: I had a chat with GPT yesterday about this after some confusions I had:

Screenshot 2023-06-27 at 09 28 46

Screenshot 2023-06-27 at 09 29 37


What do you think? Should we enforce a standard on this? Personally, I like the {Component()} syntax for sub-components as it communicates that this is not an external component more clear, but I am open to doing things either way.

tormodAase: Here is more info about `Component()` vs `<Component />`
https://dev.to/igor_bykov/react-calling-functional-components-as-functions-1d3l

packages/components/src/components/Cell/index.ts
HaakonSvane: fixed in 06e1591

packages/components/src/components/Cell/Cell.tsx
HaakonSvane: fixed in 06e1591

packages/components/src/components/Cell/Cell.tsx
HaakonSvane: I think it should be left like it is for the moment. Since the component is stateless, there is no issues calling it like this. I have created #121 to track this.

packages/components/package.json
HaakonSvane: fixed in db03b57

tormodAase (approved):

PR #128, opened 2023-07-21
fix: outlined button should now have the same height as other buttons

### LINKED ISSUE #127: Primary Button and Outlined Button does not have the same height
image
(most likely because the outlined variant has a border, while the primary variant does not)

tormodAase:
Closes #127


HaakonSvane (approved):
Looks nice. Good fix 👌
Please write a changeset for this before merging

PR #129, opened 2023-07-24
feat: Dialog component, alert function for easy usability

tormodAase:
components now export an `alert` function, that takes almost identical arguments as React Native's `Alert.alert`, with the possibility to add icons to the buttons. As long as the app uses EDSProvider, the user should be able to use our `alert` function. We also export the Dialog component, which has almost identical API as `eds-core-react`. However, it is not recommended to use that component unless `alert` does not satisfy the user's needs.

New alert function vs React native alert:
image

iPhone 8 vs iPad Mini:
image

Please test this locally and see what you think. Also, I am constantly getting this error both in main and in this branch. Don't know if this is an issue only on my machine, or something I've accidentally introduced
image


tormodAase:
As you can see in the images, the buttons have slightly different heights (outlined vs other buttons). However, that issue is resolved in [this PR](https://github.com/equinor/mad/pull/128)

packages/components/src/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,40 @@
...
+        width: "100%",
+        height: "100%",
+        justifyContent: "center",
+        alignItems: "center",
+        backgroundColor: "rgba(0,0,0,0.4)",
+    },
+    dialog: {
+        borderRadius: 4,
+        height: 213,
+        width: 500,
```
HaakonSvane, line 36: I am not too much of a fan of statically set widths and heights, but if you do, please add these to the master token instead of setting them here.
The height should most definitely be dynamic and instead have a `minHeight` property set to avoid it being too small.

packages/components/src/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,40 @@
...
+}
+
+const themeStyles = EDSStyleSheet.create((theme, props: { style: ViewStyle | undefined, dimensions: ScaledSize }) => ({
+    background: {
+        position: "absolute",
+        width: "100%",
+        height: "100%",
+        justifyContent: "center",
+        alignItems: "center",
+        backgroundColor: "rgba(0,0,0,0.4)",
```
HaakonSvane, line 31: This value should be defined in the master token for other similar backdrops we will add in the future

packages/components/src/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,40 @@
...
+    return <Portal name="scrim">
+        <Pressable style={styles.background} onPress={() => isDismissable && onClose()}>
+            <Paper elevation="aboveScrim" style={styles.dialog}>
+                {children}
+            </Paper>
+        </Pressable>
+    </Portal>
+}
+
+const themeStyles = EDSStyleSheet.create((theme, props: { style: ViewStyle | undefined, dimensions: ScaledSize }) => ({
```
HaakonSvane, line 24: Instead of passing the style as a prop here, I think it is fine to instead append this style to a style list in the paper prop:

```tsx
<Paper elevation="aboveScrim" style={[styles.dialog, props.style]}>
    {children}
</Paper>
```

packages/components/src/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,40 @@
...
+    background: {
+        position: "absolute",
+        width: "100%",
+        height: "100%",
+        justifyContent: "center",
+        alignItems: "center",
+        backgroundColor: "rgba(0,0,0,0.4)",
+    },
+    dialog: {
+        borderRadius: 4,
```
HaakonSvane, line 34: This is defined in the master token under `geometry.border.elementBorderRadius`

packages/components/src/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,40 @@
...
+            <Paper elevation="aboveScrim" style={styles.dialog}>
+                {children}
+            </Paper>
+        </Pressable>
+    </Portal>
+}
+
+const themeStyles = EDSStyleSheet.create((theme, props: { style: ViewStyle | undefined, dimensions: ScaledSize }) => ({
+    background: {
+        position: "absolute",
```
HaakonSvane, line 26: you can use `...StyleSheet.absoluteFillObject` to achieve the same here

packages/components/src/components/Dialog/DialogActions.tsx
```diff
@@ -0,0 +1,22 @@
...
+export type DialogActionsProps = PropsWithChildren<{ align?: "left" | "right" }>
+export const DialogActions = ({ align = "left", children }: DialogActionsProps) => {
+    const styles = useStyles(themeStyles, align)
+    return <View style={styles.actionsContainer}>
+        {children}
+    </View>
+}
+
+const themeStyles = EDSStyleSheet.create((theme, align: "left" | "right") => ({
+    actionsContainer: {
```
HaakonSvane, line 16: `gap` and `padding` values should use existing (or new??) values in the master token

packages/components/src/components/Dialog/DialogCustomContent.tsx
```diff
@@ -0,0 +1,21 @@
...
+    const styles = useStyles(themeStyles);
+    return <View style={styles.customContentContainer}>
+        {children}
+    </View>
+}
+
+const themeStyles = EDSStyleSheet.create(() => ({
+    customContentContainer: {
+        flex: 1,
+        padding: 16,
```
HaakonSvane, line 18: padding values should use existing (or new??) values in the master token

packages/components/src/components/Dialog/DialogHeader.tsx
```diff
@@ -0,0 +1,28 @@
...
+    const styles = useStyles(themeStyles);
+    return <View style={styles.header}>
+        <Typography variant="h6" style={styles.title} >{props.children}</Typography>
+    </View>
+
+}
+
+const themeStyles = EDSStyleSheet.create(theme => ({
+    header: {
+        height: 49,
```
HaakonSvane, line 17: Maybe this should be dynamic instead.
Seeing as this is a standalone component in contrast to a prop on the `Dialog` component, it should provide maximum flexibility and therefore allow for stuff like multiple lines of text or just any container.

packages/components/src/components/Dialog/DialogHeader.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+export const DialogHeader = (props: TextChildren) => {
+    const styles = useStyles(themeStyles);
+    return <View style={styles.header}>
+        <Typography variant="h6" style={styles.title} >{props.children}</Typography>
+    </View>
+
+}
+
+const themeStyles = EDSStyleSheet.create(theme => ({
```
HaakonSvane, line 15: magic values should reference the master token instead

packages/components/src/components/Dialog/service/DialogServiceProvider.tsx
```diff
@@ -0,0 +1,37 @@
...
+    const isDialogOpen = !!dialogContent;
+    return <Dialog isOpen={isDialogOpen} isDismissable={false}>
+        <Dialog.Header>{dialogContent?.title}</Dialog.Header>
+        <Dialog.CustomContent>
+            <Typography>{dialogContent?.message}</Typography>
+        </Dialog.CustomContent>
+        <Dialog.Actions align="right">
+            {dialogContent?.buttons.map((button) =>
+                <Button
+                    key={JSON.stringify(button)}
```
HaakonSvane, line 23: What is the reasoning for this type of keying? I just did a console.log of one of these and got

`{"text":"Dismiss","style":"destructive"}`

Meaning that we make an implicit assumption that no two buttons will be configured the same.
I know that indexes are generally discouraged, but I think it should be fine in this case seing as this list is static.

packages/components/src/components/EDSProvider/EDSProvider.tsx
```diff
@@ -24,11 +25,14 @@ export const EDSProvider = (
             <GestureHandlerRootView style={{ flex: 1 }}>
                 <PortalProvider>
                     <Portal.Host style={{ flex: 1 }} name="root">
-                        {props.children}
+                        <Portal.Host style={{ flex: 1 }} renderPortalsFirst={false} name="scrim">
```
HaakonSvane, line 28: See my comment on using `RootModal` here instead

HaakonSvane (changes requested):
Very nice shit 💯 
Check my comments for some requested changes and suggestions

packages/components/src/components/Dialog/Dialog.tsx
tormodAase: I think I'll just create a scrimcomponent instead (with values in mastertoken)

packages/components/src/components/Dialog/Dialog.tsx
tormodAase: Wouldn't I need to use StyleSheet.flatten or StyleSheet.compose in order for that to work on the web? Atleast you had to do that before.

Also, I think it's cleaner to leave styling logic outside of the jsx

packages/components/src/components/EDSProvider/EDSProvider.tsx
tormodAase: Created ScrimProvider

packages/components/src/components/Dialog/service/DialogServiceProvider.tsx
tormodAase: index seem to work

packages/components/src/components/Dialog/DialogHeader.tsx
tormodAase: I haven't seen any designs where the header is more than 50px tall (49 + border). I don't think giving the flexibility to adjust it is a good idea.

We could make title a prop instead but let's save that rewrite for later

packages/components/src/components/Dialog/DialogCustomContent.tsx
tormodAase: Fixed

packages/components/src/components/Dialog/DialogHeader.tsx
tormodAase: Added it as a value in mastertoken though

packages/components/src/components/Dialog/DialogHeader.tsx
tormodAase: Fixed

packages/components/src/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { useStyles } from "../../hooks/useStyles";
+import { Paper } from "../Paper";
+import { Portal } from "../Portal";
+
+type DialogProps = PropsWithChildren<{ isOpen: boolean, style?: ViewStyle }> & ({ isDismissable?: false, onClose?: () => void } | { isDismissable: true, onClose: () => void })
+
+export const Dialog = ({ isOpen, style, isDismissable, onClose, children }: DialogProps) => {
+    const dimensions = useWindowDimensions();
+    const styles = useStyles(themeStyles, { style, dimensions })
+    LayoutAnimation.configureNext({ ...LayoutAnimation.Presets.spring, duration: 100 })
```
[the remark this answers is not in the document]
tormodAase: Using popincontainer screwed up a lot of things, think I'll just leave it as is for now
image

packages/components/src/components/Dialog/DialogHeader.tsx
tormodAase: > I haven't seen any designs where the header is more than 50px tall (49 + border). I don't think giving the flexibility to adjust it is a good idea.

To be more clear, if a developer creates a title long enough to cover 2 lines, it should send a clear message that their title suck, and they should reduce the length of it.

tormodAase: Like this 
image

packages/components/src/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { Paper } from "../Paper";
+import { Scrim, ScrimProps } from "../Scrim";
+
+type DialogProps = Omit<ScrimProps, "onPress"> & { style?: ViewStyle } & { onScrimPress?: ScrimProps["onPress"] }
+
+export const Dialog = ({ isOpen, style, onScrimPress, children }: DialogProps) => {
+    const dimensions = useWindowDimensions();
+    const styles = useStyles(themeStyles, { style, dimensions })
+    return <Scrim isOpen={isOpen} onPress={onScrimPress}>
+        <Paper elevation="aboveScrim" style={styles.dialog}>
```
HaakonSvane, line 14: Styles are one common prop that users might find themselves trying to overwrite when using the component, but there are actually a lot of other props defined on `View` that could be handy like accessibility and touch event behaviour configuration. In essence, the Dialog component is a View wrapper since it basically consists of a `Paper` component (which is most definately a styled `View` wrapper). Because of this, I think we should define the props for the Dialog component like this:

```tsx
// Note the missing style prop:
type DialogProps = Omit<ScrimProps, "onPress"> & {onScrimPress?: ScrimProps["onPress"]}; 

// Notice the typing for the props now:
const Dialog = ({
  isOpen,
  onScrimPress,
  children,
  ...rest,
}: DialogProps & ViewProps) => {
  // .. 
  return (
    <Scrim isOpen{isOpen} onPress{onScrimPress}>
    // Note how we spread out the remaining ViewProps into the Paper component (which acceps ViewProps)
      <Paper elevation="aboveScrim" style={[styles.dialog, rest.styles]} {...rest}>
        {children}
      </Paper>
    </Scrim>
  )
}
```

packages/components/src/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,29 @@
...
+    </Scrim>
+}
+
+const themeStyles = EDSStyleSheet.create((theme, props: { style: ViewStyle | undefined, dimensions: ScaledSize }) => ({
+    dialog: {
+        backgroundColor: "white",
+        borderRadius: theme.geometry.border.elementBorderRadius,
+        minHeight: theme.geometry.dimension.dialog.minHeight,
+        width: theme.geometry.dimension.dialog.defaultWidth,
+        maxWidth: props.dimensions.width - (theme.spacing.container.paddingVertical * 2),
```
HaakonSvane, line 26: Is the `maxWidth` value needed here as you are setting the height explicitly?

packages/components/src/components/EDSProvider/EDSProvider.tsx
```diff
@@ -23,12 +25,15 @@ export const EDSProvider = (
             value={{ colorScheme: props.colorScheme, density: props.density }}>
             <GestureHandlerRootView style={{ flex: 1 }}>
                 <PortalProvider>
-                    <Portal.Host style={{ flex: 1 }} name="root">
-                        {props.children}
-                    </Portal.Host>
+                    <ScrimProvider>
```
HaakonSvane, line 28: The naming of this provider makes it seem like it is different to the `PortalProvider`.
Maybe rename this to `ScrimPortal`to better communicate its intent (?)

packages/components/src/components/Dialog/Dialog.tsx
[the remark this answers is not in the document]
HaakonSvane: I had no issue when testing when I used this implementation:

```tsx
export const Dialog = ({ isOpen, style, onScrimPress, children }: DialogProps) => {
    const dimensions = useWindowDimensions();
    const styles = useStyles(themeStyles, { style, dimensions })
    return <Scrim isOpen={isOpen} onPress={onScrimPress}>
        <PopInContainer>
            <Paper elevation="aboveScrim" style={styles.dialog}>
                {children}
            </Paper>
        </PopInContainer>
    </Scrim>
}
```

Did you try a similar solution to this?

HaakonSvane (changes requested):
Very well done 🔥 
I left some new comments / questions

packages/components/src/components/Dialog/DialogHeader.tsx
HaakonSvane: In that case, we should just drop the `Dialog.Header` component and just have `title` be a prop on the Dialog component instead. You mentioned this, but I actually think this change should be done for this PR and not the future. Why bother introducing a component that we know will be removed? It could create frustration with someone potentially implementing this component while we fix it

packages/components/src/components/Dialog/Dialog.tsx
tormodAase: Yes. By default the width will be 500. But if the screen is small enough, the width will shrink depending on the screen width

packages/components/src/components/EDSProvider/EDSProvider.tsx
tormodAase: I don't think how the provider works needs to be reflected in the name

apps/chronicles/screens/components/DialogScreen.tsx
```diff
@@ -0,0 +1,57 @@
+import { Button, EDSStyleSheet, Spacer, Typography, alert, useStyles } from "@equinor/mad-components";
+import { Alert } from "react-native";
+import { ScrollView } from "react-native-gesture-handler"
```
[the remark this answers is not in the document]
tormodAase: Both scrollviews work

HaakonSvane (approved):
Hella fresh 🔥

apps/chronicles/screens/components/DialogScreen.tsx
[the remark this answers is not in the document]
HaakonSvane: Yes, but we should stick to using the same components for all our ScrollViews unless you specifically need GH's

PR #161, opened 2023-08-12
chore(lint): create linting system

### LINKED ISSUE #140: Create a proper linting system
Our [eslint package](https://github.com/equinor/mad/tree/main/packages/eslint-config-mad-custom) is a great place to start off by setting some proper linting rules that should apply for everyone in the mobile application delivery teams at Equinor, both apps and platform.

- [x] Rename package to include equinor (i.e: `@equinor/eslint-config-something-here`). Also give it a nice name :) 
- [x] Make sure all monorepo packages and apps apply the eslint rules
- [x] Implement lint on PR as part of the acceptance criterias (use github actions. `lint fix` first before reporting any remaining issues):
- [ ] Build the config with rules we decide on

HaakonSvane:
- Reimagine our previous ESLint package
- Create basic README.md
- Add code lint and format as part of PR validation (workflow)

closes #140


HaakonSvane:
More info: https://github.com/equinor/mad/discussions/142#discussioncomment-6707547

packages/eslint-config-mad/README.md
```diff
@@ -0,0 +1,29 @@
...
+This package includes all ESLint configurations used in the MAD Team at Equinor.
+
+## Config structure
+
+The main package consists of multiple entrypoints:
+
+-   `base`
+-   `react`
+-   `typescript`
+
```
tormodAase, line 12: Looks like a good start! So if I understand it correctly, `base` rules that fit any javascript project, `typescript` has rules that fit any typescript project, and `react` any react project, correct?
And if you use `@equinor/mad`, you use all rules?

I would suggest having one for react-native as well. Should we create an issue for it?

tormodAase: And I guess the goal is to have two sets of rules that our projects would use: `@equinor/mad/web`, which extends `base`, `react`, `typescript` and maybe a web-specific ruleset, and `@equinor/mad/mobile` which would extend `base`, `react`, `typescript` and some mobile specific ruleset (react-native). Am I correct?

HaakonSvane: That is correct.
This should be a good starting point for our projects I think. 
As we keep on arguing over details in PR's, I think we should make feature requests to this package and let it grow organically. 
As time passes, I hope the best structure will emerge.

tormodAase (approved):

PR #176, opened 2023-08-30
feat!: environment provider

tormodAase:
This PR adds environment provider, which means the environment banner component no longer needs an environment prop.


HaakonSvane (approved):
Very nice very nice <3

PR #182, opened 2023-09-04
fix: navigation package dependency issues

HaakonSvane:
- set `@equinor/mad-components` to normal dependency in `@equinor/mad-navigation`.
- reinstall. 
- export `EnvironmentProvider` through `@equinor/mad-navigation`
- set Chronicles to use newly exported provider


HavardNot (approved):
LGTM

PR #183, opened 2023-09-04
feat: add switch component

Sandtory:
- add the main Switch component, based off of Button component.
- employ React's forwardRef to obtain a reference to the underlying view.
- integrate custom hooks (useStyles, useToken) to derive theme-specific styles and animation timings.
- incorporated Paper component for the knob, adding depth with elevation.
- add sub component SmallSwitch with typeOf Switch.
- add SwitchScreen with examples of Switch usage.

 ^ written by HenningGPT

### DIFF
--- a/.changeset/thick-papayas-promise.md
+++ b/.changeset/thick-papayas-promise.md
@@ -2,4 +2,4 @@
 "@equinor/mad-components": patch
 ---
 
-Added new component Switch, child component SmallSwitch and a SwitchScreen
+Added `Switch` and `Switch.Small` components
--- a/.changeset/thick-papayas-promise.md
+++ b/.changeset/thick-papayas-promise.md
@@ -1,0 +1,5 @@
+---
+"@equinor/mad-components": patch
+---
+
+Added new component Switch, child component SmallSwitch and a SwitchScreen
--- a/apps/chronicles/screens/components/SwitchScreen.tsx
+++ b/apps/chronicles/screens/components/SwitchScreen.tsx
@@ -1,7 +1,6 @@
-import { Switch } from "../../../../packages/components/src/components/Switch/Switch";
 import React, { useState } from "react";
 import { ScrollView, View } from "react-native";
-import { EDSStyleSheet, Spacer, Typography, useStyles } from "@equinor/mad-components";
+import { EDSStyleSheet, Spacer, Typography, useStyles, Switch } from "@equinor/mad-components";
 
 export const SwitchScreen = () => {
     const styles = useStyles(themeStyles);
--- a/apps/chronicles/screens/components/SwitchScreen.tsx
+++ b/apps/chronicles/screens/components/SwitchScreen.tsx
@@ -1,13 +1,13 @@
 import React, { useState } from "react";
 import { ScrollView, View } from "react-native";
-import { EDSStyleSheet, Spacer, Typography, useStyles } from "@equinor/mad-components";
-import { Switch } from "../../../../packages/components/src/components/Switch/";
+import { EDSStyleSheet, Spacer, Typography, useStyles, Switch } from "@equinor/mad-components";
 
 export const SwitchScreen = () => {
     const styles = useStyles(themeStyles);
     const [isActiveSwitch1, setIsActiveSwitch1] = useState(false);
     const [isActiveSwitch2, setIsActiveSwitch2] = useState(true);
     const [isActiveSwitchSmall, setIsActiveSwitchSmall] = useState(false);
+    const [isDisabledSwitchSmall, setIsDisabledSwitchSmall] = useState(false);
     const [isDisabled, setIsDisabled] = useState(true);
     const [isDisabledActiveSwitch, setDisabledActiveSwitch] = useState(true);
 
@@ -23,6 +23,10 @@ export const SwitchScreen = () => {
         setIsActiveSwitchSmall(newState);
     };
 
+    const handleSmallSwitchDisabled = (newState: boolean) => {
+        setIsDisabledSwitchSmall(newState);
+    };
+
     const handleDisabledActiveState = (newState: boolean) => {
         setIsDisabled(newState);
         setDisabledActiveSwitch(newState);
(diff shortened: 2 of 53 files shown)


apps/chronicles/screens/components/SwitchScreen.tsx
```diff
@@ -0,0 +1,101 @@
+import React, { useState } from "react";
+import { ScrollView, View } from "react-native";
+import { EDSStyleSheet, Spacer, Typography, useStyles } from "@equinor/mad-components";
+import { Switch } from "../../../../packages/components/src/components/Switch/";
```
HaakonSvane, line 4: This import should be pointing to the package and not relative reference in the monorepo.

HaakonSvane: It is probably why we dont get dark mode support btw!

apps/chronicles/screens/components/SwitchScreen.tsx
```diff
@@ -0,0 +1,101 @@
...
+        </ScrollView>
+    );
+};
+
+const themeStyles = EDSStyleSheet.create(theme => ({
+    container: {
+        paddingHorizontal: theme.spacing.container.paddingHorizontal,
+        paddingVertical: theme.spacing.container.paddingVertical,
+    },
+    SwitchRow: {
```
HaakonSvane, line 89: style names should not be capitalized

apps/chronicles/screens/components/SwitchScreen.tsx
```diff
@@ -0,0 +1,101 @@
...
+        paddingVertical: 8,
```
HaakonSvane, line 90: Instead of hard coding this value, if you want spacing between your components, try to use the `<Spacer />` component in mad components. It comes with some predefined values

apps/chronicles/screens/components/SwitchScreen.tsx
```diff
@@ -0,0 +1,101 @@
...
+        flexDirection: "row",
+        justifyContent: "space-around",
+    },
+    switchDescription: {
+        flexDirection: "column",
+        alignItems: "center",
+        paddingHorizontal: 8,
```
HaakonSvane, line 97: Same here as for my other comment on spacing

apps/chronicles/types.tsx
```diff
@@ -17,6 +17,8 @@ export type DiscoverStackParamList = {
     Discover: undefined;
     Paper: undefined;
     Button: undefined;
+    Switch: undefined;
+    SmallSwitch: undefined;
```
HaakonSvane, line 21: We don't need a separate navigation route for the small switch since it is shown on the same screen as the big one

apps/chronicles/screens/components/SwitchScreen.tsx
```diff
@@ -0,0 +1,101 @@
...
+                    <Switch
+                        active={isDisabledActiveSwitch}
+                        disabled={isDisabled}
+                        onChange={handleDisabledActiveState}
+                    />
+                    <Typography>{"You can't turn me off"}</Typography>
+                </View>
+            </View>
+            <Spacer />
+            <Typography variant="h2">Small</Typography>
```
HaakonSvane, line 67: does the small Switch support disabled? If so, maybe show it here as an example as well

packages/components/src/components/Switch/Switch.tsx
```diff
@@ -0,0 +1,202 @@
...
+        const inactiveBackgroundAnimation = Animated.timing(backgroundProgressValue, {
+            toValue: 0,
+            duration: token.timing.animation.slow,
+            useNativeDriver: true,
+            easing: Easing.inOut(Easing.ease),
+        });
+
+        const backgroundDisplacement = backgroundProgressValue.interpolate({
+            inputRange: [0, 1],
+            outputRange: [0.001, 1],
```
HaakonSvane, line 82: Why was this value set to 0.001 ?

packages/components/src/components/Switch/Switch.tsx
```diff
@@ -0,0 +1,202 @@
...
+            easing: Easing.inOut(Easing.ease),
+        });
+
+        const displacement = knobProgressValue.interpolate({
+            inputRange: [0, 1],
+            outputRange: [0, WIDTH - KNOB_SIZE * 1.7],
+            extrapolate: "clamp",
+        });
+
+        const activeBackgroundAnimation = Animated.timing(backgroundProgressValue, {
```
HaakonSvane, line 66: Notice that this animation does exactly the same as the activeKnobAnimation does.
This means we dont need two animations (one for the knob and one for the background). Since the animated values that the knob needs are different than what the background needs, this is handled in the interpolate methods you have defined.

By removing activeBackgroundAnimation and inactiveBackgroundAnimation, you can reduce the code in this file by a good chunk 👍

packages/components/src/components/Switch/Switch.tsx
```diff
@@ -0,0 +1,202 @@
...
+    return {
+        toggleContainer: {
+            flexDirection: "row",
+            alignItems: "center",
+        },
+        pressableContainer: {
+            height: HEIGHT,
+            width: WIDTH,
+            padding: 7,
+            borderRadius: 100,
```
HaakonSvane, line 169: Just to safe, maybe set the border radius to be WIDTH / 2 so that it scales with size

.changeset/thick-papayas-promise.md
```diff
@@ -0,0 +1,5 @@
+---
+"@equinor/mad-components": patch
+---
+
+Added new component Switch, child component SmallSwitch and a SwitchScreen
```
HaakonSvane, line 5: I would rephrase this to
> Added `Switch` and `Switch.Small` components

the code blocks helps users understand that these are actual jsx tags that can be used.
Also, the changeset is written for the mad components package, so we don't need to mention any SwitchScreens since that belongs to the chronicles app.

HaakonSvane (changes requested):
Super awesome. Great work. You should be really proud 🔥
I added some minor comments for things we should look into / correct before we merge. other than that, quality code dude 😎

apps/chronicles/screens/components/SwitchScreen.tsx
Sandtory: this was taken from the button component, should i change it there aswell?

packages/components/src/components/Switch/Switch.tsx
Sandtory: it was based on some old version where the value had to be greater than zero, it will be changed

HaakonSvane (approved):
Awesome job 👍

PR #194, opened 2023-09-12
fix: `Cell` styling not working correctly outside of scroll views

HaakonSvane:
- Remove full flex in `Cell` content container styling


Sandtory (approved):
You really know your way around the cell, good work!

PR #212, opened 2023-10-03
fix(search): correct search layout

Jonas-Omdal:
- Using the search feature in another projects did not work well; the search bar became flat.
- Adjusted layout


HaakonSvane (approved):
👍

HaakonSvane (approved):
Good job 👍

PR #215, opened 2023-10-05
feat: `HeaderHeightProvider`

tormodAase:


packages/navigation/src/header-height-context/headerHeightContext.tsx
HaakonSvane, line 1: If there is only one file inside the header-height-context folder (besides the index file of course), should the folder be omitted? You could remove a lot of clutter by just having the `headerHeightContext` file under src

packages/navigation/src/header-height-context/headerHeightContext.tsx
```diff
@@ -0,0 +1,61 @@
...
+        <HeaderHeightContext.Provider value={contextValues}>
+            {children}
+        </HeaderHeightContext.Provider>
+    );
+};
+
+/**
+ * Used within mad navigation to provide the header height
+ */
+const _useMadNavigationHeaderHeight = () => {
```
HaakonSvane, line 35: What is the intention of using underscore here?
It might be easier to just get this value straight out from the context instead of creating this wrapper.

i.e, using it in HeightSender as
```ts
export const HeightSender = () => {
  const { setHeight } = useContext(HeaderHeightContext);
  // ...
}
```

packages/navigation/src/header-height-context/headerHeightContext.tsx
```diff
@@ -0,0 +1,61 @@
...
+ */
+export const useHeaderHeight = () => {
+    const context = useContext(HeaderHeightContext);
+    return context.height;
+};
+
+/**
+ * Used within mad navigation to provide the header height
+ */
+export const HeightSender = () => {
```
HaakonSvane, line 52: Is there a reason this can't be a hook instead of a component?

HaakonSvane (approved):
Good ridin' cowboy 🐮 🐄 🤠 Yehaaaaw

packages/navigation/src/header-height-context/headerHeightContext.tsx
tormodAase: This is true! I normally create a hook for using the context for simplicity, and to make it possibly to hook into the context without exporting the context itself. However since HeightSender is defined in the same file we don't need the middleman. Will fix this!

packages/navigation/src/header-height-context/headerHeightContext.tsx
tormodAase: Yes! We add this to the custom render function of the decorator. Since the render function is a render function and not a component, we can't use hooks inside it.

HaakonSvane: Ahh I see 👍

packages/navigation/src/header-height-context/headerHeightContext.tsx
tormodAase: I have no strong views on this, but in my opinion it looks better to have the different features sorted into folders. It's easier to find the index file that way since folders come before files, instead of index being mixed in between files.
image

vs 

image

But I'm down for both solutions. Another upside of sorting into folders is that you have made the features ready to be expanded on. If we later on decide to expand on NavigationContainer or HeaderHeightProvider and need more files, we would have to put them back into folders.

Anyway up to you which version you prefer

HaakonSvane (approved):
👌

PR #220, opened 2023-10-09
feat: add autocomplete

Sandtory:
Add autocomplete component.
Changed inputprops.

### DIFF
--- a/apps/chronicles/screens/components/AutocompleteScreen.tsx
+++ b/apps/chronicles/screens/components/AutocompleteScreen.tsx
@@ -78,7 +78,7 @@ export const AutocompleteScreen = () => {
                     label="Unearth your Ground Critters"
                     placeholder="Who's scuttling below?"
                     onSelect={selectedOption => console.log(`Unearthed: ${selectedOption}`)}
-                    type="multi-select"
+                    multiple
                 />
             </View>
         </ScrollView>
--- a/apps/chronicles/screens/components/AutocompleteScreen.tsx
+++ b/apps/chronicles/screens/components/AutocompleteScreen.tsx
@@ -43,22 +43,6 @@ const awesomeAnimals = [
 
 export const AutocompleteScreen = () => {
     const styles = useStyles(themedStyles);
-    const [selectedToggleButtons, setSelectedToggleButtons] = useState<string[]>([]);
-    const handleToggle = (animal: string) => {
-        if (selectedToggleButtons.includes(animal)) {
-            // If the animal is already selected, remove it from the array
-            setSelectedToggleButtons(prevState => prevState.filter(item => item !== animal));
-        } else {
-            // If the animal is not selected, add it to the array
-            setSelectedToggleButtons(prevState => [...prevState, animal]);
-        }
-    };
-    const toggleOptions = awesomeAnimals.map(animal => ({
-        title: animal,
-        active: selectedToggleButtons.includes(animal),
-        iconName: selectedToggleButtons.includes(animal) ? "radiobox-marked" : "radiobox-blank",
-        onPress: () => handleToggle(animal),
-    }));
 
     return (
         <ScrollView contentInsetAdjustmentBehavior="automatic" contentContainerStyle={{ flex: 1 }}>
@@ -94,14 +78,8 @@ export const AutocompleteScreen = () => {
                     label="Unearth your Ground Critters"
                     placeholder="Who's scuttling below?"
                     onSelect={selectedOption => console.log(`Unearthed: ${selectedOption}`)}
-                    toggleOptions={toggleOptions}
+                    type="multi-select"
                 />
-            </View>
-            <View style={styles.container}>
-                <Typography variant="h3">Selected Animals:</Typography>
-                {selectedToggleButtons.map(animal => (
-                    <Typography key={animal}>{animal}</Typography>
-                ))}
             </View>
         </ScrollView>
     );
--- a/apps/chronicles/screens/components/AutocompleteScreen.tsx
+++ b/apps/chronicles/screens/components/AutocompleteScreen.tsx
@@ -55,7 +55,7 @@ export const AutocompleteScreen = () => {
                     options={awesomeAnimals}
                     label="Click me and spot a Jungle Creature"
                     placeholder="Type a name and see who's lurking..."
-                    onSelect={selectedOption => console.log(`Spotted: ${selectedOption}`)}
+                    onSelect={() => null}
                 />
             </View>
             <View style={styles.container}>
@@ -65,7 +65,7 @@ export const AutocompleteScreen = () => {
                     options={awesomeAnimals}
                     label="Find a Tree-dweller"
                     placeholder="Who's swinging in the trees?"
-                    onSelect={selectedOption => console.log(`Found: ${selectedOption}`)}
+                    onSelect={() => null}
                 />
             </View>
             <View style={styles.container}>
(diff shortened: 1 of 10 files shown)


packages/components/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,221 @@
...
+            )}
+        </View>
+    );
+};
+
+const themedStyles = EDSStyleSheet.create(
+    (theme, { inputLayout }: { inputLayout?: LayoutRectangle }) => ({
+        container: {
+            // styles for the container of the component
+        },
```
tormodAase, line 196: This style can probably be removed since it does not contain anything

packages/components/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,221 @@
...
+            alignItems: "center",
+        },
+        closeIcon: {
+            marginLeft: 8,
+        },
+        menuIcon: {
+            marginRight: 3,
+        },
+    }),
+);
```
tormodAase, line 221: it looks like `option` and `optionText` is not being used. Those styles can be removed

packages/components/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,221 @@
...
+    };
+    const handleClearSelectedOptions = () => {
+        setSelectedToggleOptions([]);
+    };
+
+    useEffect(() => {
+        setFilteredOptions(
+            options.filter(option => option.toLowerCase().includes(inputValue.toLowerCase())),
+        );
+    }, [inputValue]);
```
tormodAase, line 82: I don't think it would hurt to add `options` as a dependency here, but if you decide not to, I would disable the eslint rule.

packages/components/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,221 @@
...
+        handleMenuClose();
+    };
+    const handleToggleOptionPress = (optionTitle: string) => {
+        if (selectedToggleOptions.includes(optionTitle)) {
+            setSelectedToggleOptions(prev => prev.filter(title => title !== optionTitle));
+        } else {
+            setSelectedToggleOptions(prev => [...prev, optionTitle]);
+        }
+    };
+    const handleClearSelectedOptions = () => {
```
tormodAase, line 74: Apparently this is never used

packages/components/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,221 @@
...
+        setSelectedToggleOptions([]);
+    };
+
+    useEffect(() => {
+        setFilteredOptions(
+            options.filter(option => option.toLowerCase().includes(inputValue.toLowerCase())),
+        );
```
tormodAase, line 81: It looks like `filteredOptions` will always have the same state given the same options, and the same inputValue. This means you don't have to make it a state! You can get rid of this useEffect and instead use `useMemo`.

Here is an isolated example:
```tsx
type Props = {options: string[]}
const Component = ({options}: Props) => {
  const [inputValue, setInputValue] = useState('');
  const filteredOptions = useMemo(() => options.filter(option => option.toLowerCase().includes(inputValue.toLowerCase())), [options, inputValue])
  //rest of the code here
}
```

Also, keep in mind that the `filteredList` will never filter `toggleOptions`, which looks like a bug to me

apps/chronicles/screens/components/AutocompleteScreen.tsx
```diff
@@ -0,0 +1,116 @@
...
+            // If the animal is not selected, add it to the array
+            setSelectedToggleButtons(prevState => [...prevState, animal]);
+        }
+    };
+    const toggleOptions = awesomeAnimals.map(animal => ({
+        title: animal,
+        active: selectedToggleButtons.includes(animal),
+        iconName: selectedToggleButtons.includes(animal) ? "radiobox-marked" : "radiobox-blank",
+        onPress: () => handleToggle(animal),
+    }));
```
tormodAase, line 61: To me it's a bit strange that the parent component has to handle the state for which options are toggled. Maybe we can look at this together to see how to improve it

packages/components/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,221 @@
...
+                                onPress={() => {
+                                    setInputValue(item);
+                                    onSelect(item);
+                                    setIsOptionsVisible(false);
+                                }}
+                            >
+                                <Text style={styles.optionText}>{item}</Text>
+                            </TouchableOpacity>
+                        )}
+                    /> */}
```
tormodAase, line 185: This component is getting quite long. Consider separating it into multiple components

tormodAase (changes requested):
Good job 👊🏻! Please check my comments! We can also look at some of these parts together if you want!

apps/chronicles/screens/components/AutocompleteScreen.tsx
```diff
@@ -0,0 +1,94 @@
...
+import { ScrollView, View } from "react-native";
+import {
+    EDSStyleSheet,
+    useStyles,
+    Typography,
+    Spacer,
+    Autocomplete,
+} from "@equinor/mad-components";
+
+const awesomeAnimals = [
```
HaakonSvane, line 11: Cool animal names! 🐯 🦁 🐱 😠 🐶

apps/chronicles/screens/components/AutocompleteScreen.tsx
```diff
@@ -0,0 +1,94 @@
...
+                />
+            </View>
+            <View style={styles.container}>
+                <Typography>Explore the Jungle Canopy</Typography>
+                <Spacer />
+                <Autocomplete
+                    options={awesomeAnimals}
+                    label="Find a Tree-dweller"
+                    placeholder="Who's swinging in the trees?"
+                    onSelect={selectedOption => console.log(`Found: ${selectedOption}`)}
```
HaakonSvane, line 68: We should perhaps remove these console logs. When using the app, you don't see them. If we want to show the user the selections, consider showing it in the app instead. I'm ok with not showing anything as well though. Whatever you want

packages/components/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,206 @@
...
+     */
+    label?: string;
+    /**
+     * A placeholder for the input field.
+     */
+    placeholder?: string;
+} & InputProps &
+    (SingleSelect | MultiSelect);
+
+export const Autocomplete: React.FC<AutocompleteProps> = ({
```
HaakonSvane, line 42: We don't need the React.FC wrapper around the types. This makes it so that the component accepts children, which I guess this does not. Look at the other types for inspiration.

packages/components/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,206 @@
...
+                                      selectedToggleOptions.includes(option),
+                                  ),
+                        )}
+                    </ScrollView>
+                </Menu>
+            )}
+        </View>
+    );
+};
+
```
HaakonSvane, line 188: Add a Autocomplete.displayName = "Autocomplete" under the component definition. See the other components in the library for more info.

HaakonSvane (approved):
```
   .------\ /------.
   |       -       |
   |               |
   |               |
   |               |
_______________________
===========.===========
  / ~~~~~     ~~~~~ \
 /|     |     |\             Request for pull has been accepted!
 W   ---  / \  ---   W     /
 \.      |o o|      ./
  |                 |
  \    #########    /
   \  ## ----- ##  /
    \##         ##/
     \_____v_____/
     
```

tormodAase:
Cannot approve for some reason

HaakonSvane (approved):
👍

PR #223, opened 2023-10-10
fix(search): fix clear text button

Jonas-Omdal:
- Fixed clear text button. Was some issues with onchange when clearing text in search.

### DIFF
--- a/packages/components/src/components/Search/Search.tsx
+++ b/packages/components/src/components/Search/Search.tsx
@@ -105,7 +105,6 @@ const themedStyles = EDSStyleSheet.create(theme => ({
     container: {
         flexDirection: "row",
         alignItems: "flex-end",
-        gap: theme.spacing.element.paddingHorizontal
     },
     adornment: {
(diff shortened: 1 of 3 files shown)


packages/components/src/components/Search/Search.tsx
```diff
@@ -108,6 +115,7 @@ export const Search = ({ cancellable, onCancelPress, onChange, ...restProps }: S
 
 const themedStyles = EDSStyleSheet.create(theme => ({
     container: {
+        flex: 1,
```
tormodAase, line 118: I would create a changeset for this fix too. it's possible to create multiple changesets per PR.

Also, it's nice if the changeset describes the bug you fixed. Looking at the current changeset, I can't really tell which issue you fixed related to onChange/onBlur

tormodAase (changes requested):
Looks good! I would update the changeset(s) to be more descriptive. Which bug(s) did you fix?

PR #231, opened 2023-10-12
fix(iconbutton): Adjust pressable container size on icon button

Jonas-Omdal:
Pressable container size was incorrect
- Changed the size on the pressable container around the icon button to it's correct size


HaakonSvane (approved):
👍

PR #238, opened 2023-10-13
feat: add scopes prop to login button

tormodAase:
Dont know if this will help with our login issues, but it shouldn't make things worse!

I've sent an admin request consent for mad api prod, so we will have to check on monday if it makes things better.

This feature adds the ability to add scopes to the loginbutton. If you don't use the new prop, the scopes will be an empty array by default


HaakonSvane (approved):
Good shit 👌👌

PR #240, opened 2023-10-19
fix: fixed autocomplete styling

Sandtory:
added flexGrow to `Autocomplete` `View` so that it is scalable


HaakonSvane (approved):
Looks very nicely put together by you

PR #249, opened 2023-10-26
fix: `.d.ts` files are now created using tsup instead of tsc

tormodAase:
This will help us with pull request validation, and make the tsup configs simpler


tormodAase:
Looks like PR validation works 😄

tormodAase:
Do we need changeset for this change?

HaakonSvane:
I dont think so, as long as it seems to generate the same types

HaakonSvane (approved):
Good stuff 👌👌❤️

PR #255, opened 2023-10-27
chore: fix PR validation

tormodAase:
Fixed by changing back to tsc for compiling declaration files.
- In watch mode, we run `tsc` as part of tsup, in the onSuccess function
- When we run `build` (which is not in watch mode), we run tsc after tsup is finished, to make sure the step throws an error if it fails


HaakonSvane (approved):
👌👌

PR #256, opened 2023-10-27
refactor!: autocomplete move multiselect

Sandtory:
change `Autocomplete` to have sub-component `Autocomplete.Multiselect`
Move state out of component to be a controlled by parent component


.changeset/olive-lamps-unite.md
```diff
@@ -0,0 +1,5 @@
+---
+"@equinor/mad-components": patch
+---
+
+changed autocomplete to have sub component multiselect
```
tormodAase, line 5: I believe this should be a minor change, since the API for the `Autoselect` component has breaking changes

tormodAase:
Also, since this PR has some breaking changes, maybe it should be considered a feat/fix? I don't know how changelogs work but it might not show up in the changelogs unless it's a feat or fix. But @HaakonSvane knows more about this

HaakonSvane:
This PR title should be `refactor!` (note the exclamation mark) to communicate a breaking change.
As for the changeset itself, we should add text to it stating that you will now have to move from using the multiselect prop to using the Autocomplete.Multiselect component instead

packages/components/src/components/Autocomplete/index.ts
```diff
@@ -1 +1,11 @@
-export * from "./Autocomplete";
+import { Autocomplete as _Autocomplete } from "./Autocomplete";
+import { MultiSelectAutocomplete } from "./MultiSelectAutocomplete";
+
+type AutocompleteFamily = typeof _Autocomplete & {
+    MultiSelect: typeof MultiSelectAutocomplete;
+};
+
+const Autocomplete = _Autocomplete as AutocompleteFamily;
+Autocomplete.MultiSelect = MultiSelectAutocomplete;
```
HaakonSvane, line 9: I think we should rename this to be `Autocomplete.Multiselect` (no capitalization of "s") since it seems to the be written in one word

HaakonSvane (approved):
🪨 📈 
You have pushed his rock with relentless resolve, so have you heaved lines of code towards the summits of our project's ambition. Your labor, far from futile, threads the fabric of a tapestry rich with purpose, each commit a testament to the notion that in the kiln of effort, we forge a life of significance. The rock of Sisyphus is not a burden, but a craft, and your craftsmanship endows our collective journey with meaning. In this pull request, I see not just a task completed, but the very essence of a life well coded.

PR #263, opened 2023-11-02
feat: settings screen

tormodAase:
This PR implements Settings Screen into mad-core

### DIFF
--- a/apps/chronicles/mad.config.ts
+++ b/apps/chronicles/mad.config.ts
@@ -17,4 +17,39 @@ export const config: MadConfig = {
         title: "Chronicles",
         logo: Logo,
     },
+    about: {
+        endpoints: [],
+        buildNumber: "1",
+    },
+    serviceNow: {
+        whatever: "",
+    },
+    settings: [
+        {
+            title: "test",
+            items: [
+                {
+                    name: "navigation",
+                    description: "With description",
+                    iconName: "ab-testing",
+                    onPress: () => undefined,
+                    title: "Navigation test",
+                },
+                {
+                    name: "button",
+                    iconName: "ab-testing",
+                    onPress: () => undefined,
+                    title: "Button test",
+                    color: "danger",
+                },
+                {
+                    name: "switch",
+                    title: "Switch test",
+                    iconName: "ab-testing",
+                    isActive: true,
+                    onChange: () => undefined,
+                },
+            ],
+        },
+    ],
 };
--- a/apps/chronicles/mad.config.ts
+++ b/apps/chronicles/mad.config.ts
@@ -6,7 +6,10 @@ export const config: MadConfig = {
     servicePortalName: "Chronicles",
     environment: "test",
     language: {
-        supportedLanguages: [{ code: "en", name: "English" }],
+        supportedLanguages: [
+            { code: "en", name: "English" },
+            { code: "nb", name: "Norwegian" },
+        ],
     },
     authentication: {
         redirectUri: "msauth.com.equinor.mad.chronicles://auth",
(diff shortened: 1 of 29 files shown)


apps/chronicles/mad.config.ts
```diff
@@ -17,4 +20,11 @@ export const config: MadConfig = {
         title: "Chronicles",
         logo: Logo,
     },
+    serviceNow: {
+        whatever: "",
```
HavardNot, line 24: This looks like a placeholder? Maybe add a todo if you are unable to implement this at the moment?

apps/chronicles/navigation/SettingsScreen.tsx
```diff
@@ -0,0 +1,33 @@
...
+                {
+                    name: "switch",
+                    title: "Test",
+                    onChange: () => undefined,
+                    isActive: true,
+                    iconName: "abacus",
+                },
+            ],
+        },
+    ];
```
HavardNot, line 30: Is this a template? Should maybe clarify it, e.g. rename to "templateConfig" or something

packages/core/src/components/screens/settings/types.tsx
```diff
@@ -0,0 +1,38 @@
...
+          title: string;
+          isActive: boolean;
+          onChange: (isActive: boolean) => void;
+          switchSize?: "small" | "normal";
+          disabled?: boolean;
+          description?: string;
+          color?: SwitchColor;
+          iconName: IconName;
+      }
+    | { name: "custom"; title: string; iconName: IconName; param: unknown };
```
HavardNot, line 31: These types could maybe be unified with the type of the component it describes. E.g. "switch" should use the type of the Switch component props, "button" the Button component props, and so on.

packages/core/src/types.ts
```diff
@@ -59,6 +59,20 @@ export type MadConfig = {
...
+         * Endpoints used by the app
+         */
+        endpoints: string[];
+        /**
+         * Build number of the app.
+         */
+        buildNumber: string;
+    };
+    serviceNow?: {
+        //TODO
```
HavardNot, line 73: Nice todo bro 😎

packages/core/tsconfig.json
```diff
@@ -8,7 +8,8 @@
         "noPropertyAccessFromIndexSignature": true,
         "noImplicitReturns": true,
         "noFallthroughCasesInSwitch": true,
-        "exactOptionalPropertyTypes": true,
+        //"exactOptionalPropertyTypes": true,
```
HavardNot, line 11: Is this temporarily commented out or can it be removed?

HavardNot (changes requested):
Looks great! I have a few QoL requested changes, but other than that we should be good to go! 🎉

apps/chronicles/mad.config.ts
HavardNot: Edit: I noticed you added a todo to the type declaration so looks like youre covered 👍

packages/core/tsconfig.json
tormodAase: Removed, hard to use mad-components with that setting enabled, when mad-components does not have it enabled

packages/core/src/components/screens/settings/types.tsx
tormodAase: Fixed

apps/chronicles/navigation/SettingsScreen.tsx
tormodAase: Fixed

HavardNot (approved):
Looks hella fresh 🧔🏻‍♂️

PR #269, opened 2023-11-15
feat: whats new screen

HavardNot:
finished implementing the WhatsNewScreen according to spec requirements for functionality and design. The "Changelog" component can be reused to display all release notes in the tbd ReleaseNotesScreen


tormodAase (commented):
Looking good! Will do a full review later!

We need some changesets. Run pnpm changesets to add changesets. I would add two changesets: one for mad-navigation and one for mad-core.  In order to accomplish this, run the command twice and select one package at a time

tormodAase (approved):

PR #272, opened 2023-11-16
fix: add JSX Elements to child props of Typography

HavardNot:
Adds support for nesting Typography components and other JSX elements. Tested it a bit in Code of Conduct and seems to handle most props with ease (e.g. View, TouchableOpacity etc.)


tormodAase (approved):

HaakonSvane (approved):
🔥

PR #276, opened 2023-11-21
fix(button): pass `style` prop to outermost component

### LINKED ISSUE #170: `Button`: `style` prop is added to nested `View`
### Package / App

MAD Components

### Version

0.4.0

### What happened?

When adding `align-self: "flex-end"` to the `style` prop on a `Button` it doesn't work as intended since these styles are added to a nested `View`. For these styles to work correctly the `style` prop needs to be added to the outermost component in `Button`.

### Other info

```TSX
// Button.tsx
...
 return (
            <View> // This `View` should probably not be here
                <View ref={ref} style={[styles.colorContainer, rest.style]}>
                    <PressableHighlight
                        disabled={disabled}
                        onPress={isToggleButton ? toggleData.toggle : onPress}
                        style={styles.pressableContainer}
                    >
                        <View style={styles.labelContainer}>
                            {loading ?
                                <DotProgress
                                    color={(disabled || variant !== "contained") ? "primary" : "neutral"}
                                    size={12} />
                                :
                                ButtonContent()}
                        </View>
                    </PressableHighlight>
                </View>
            </View>
        );
```

### Relevant log output

_No response_

dansvend:
- Remove wrapper `View` to make the `style` prop on `Button` behave as expected.
- Add `ButtonSpecificProps` type which has the same types as `ButtonProps` had.
- Update `ButtonProps` type to be `ButtonSpecificProps`  and `ViewProps`.

Closes #170

### DIFF
--- a/.changeset/thick-fireants-melt.md
+++ b/.changeset/thick-fireants-melt.md
@@ -1,0 +1,7 @@
+---
+"@equinor/mad-components": patch
+---
+
+`Button`: The `style` prop is now passed to the outermost component to make it behave as expected.
+We now export `ButtonSpecificProps` which was previously named `ButtonProps`. `ButtonProps` is still
+exported and consists of the types `ButtonSpecificProps` and `ViewProps`.
--- a/packages/components/src/components/Button/Button.tsx
+++ b/packages/components/src/components/Button/Button.tsx
@@ -10,7 +10,7 @@ import { Icon, IconName } from "../Icon";
 import { DotProgress } from "../ProgressIndicator";
 import { getBackgroundColorForButton } from "../../utils/getBackgroundColorForButton";
 
-export type ButtonProps = {
+export type ButtonSpecificProps = {
     /**
      * Label text of the button.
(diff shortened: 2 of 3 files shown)


HaakonSvane (changes requested):
Good stuff. Nice to see this cleaned out. Buttons seem to behave the same as before which is nice. Is this a breaking change? Nonetheless, I think we should add a changeset explaining that we now export the full button props.

dansvend:
> Good stuff. Nice to see this cleaned out. Buttons seem to behave the same as before which is nice. Is this a breaking change? Nonetheless, I think we should add a changeset explaining that we now export the full button props.

I believe it's not a breaking change, but I worked around this bug by wrapping a `Button` in a `View`, so there's probably some project code which can be removed after this fix.

I agree - I'll add a changeset 💪

tormodAase (approved):

HaakonSvane (approved):
Really nice. Helpful changeset 📖

PR #284, opened 2023-11-24
chore: clean:all

HavardNot:
cleans all packages and apps respective node_modules and dist/ios folder


tormodAase (dismissed):
Lookin good 🔥

.changeset/unlucky-crabs-build.md
HaakonSvane, line 1: I don't think we need to create changesets for these kinds of changes. The changesets are the changelog-items that end up on users release notes for the packages. This change feels to me more like a developer tool for the frontend tigers working in the monorepo. Should we remove this?

package.json
```diff
@@ -32,7 +32,8 @@
         "dev:all": "turbo run dev",
         "lint:all": "turbo run lint --continue --",
         "format:all": "prettier --write .",
-        "publish:all": "turbo run build && changeset version && changeset publish"
+        "publish:all": "turbo run build && changeset version && changeset publish",
+        "clean:all": "turbo run clean"
```
HaakonSvane, line 36: Should this also delete `node_modules` in the root directory?

turbo.json
```diff
@@ -16,6 +16,7 @@
         "generate": {
             "cache": false
         },
-        "test": {}
+        "test": {},
+        "clean": {}
```
HaakonSvane, line 20: Turbo's caching system seem to sometimes not want to delete `node_modules` after being ran once. I created a `node_modules` folder in one of the packages and ran the script again, but turbo cached it all and did not want to delete it. We should disable caching by adding a 
```json
"cache": false
```
field to this job

HaakonSvane (dismissed):
Great job frontend dancer 👯 
It seems the maintenance api wrapper package was left out. Perhaps we should add the clean script to that as well

.changeset/unlucky-crabs-build.md
HavardNot: removed

package.json
HavardNot: done

turbo.json
HavardNot: done

HaakonSvane (approved):
👍

PR #310, opened 2023-12-13
feat: Add `AboutScreen`

### LINKED ISSUE #235: About Screen
### Package / App

Other

### Description

We need a common "about screen" for our apps
Design: https://www.figma.com/file/GmYi1qTwOkAM4ZyK5Bsj31/Component-library?type=design&node-id=4-22998&mode=dev

### Alternatives

_No response_

### Other info

_No response_

dansvend:
Closes #235

### DIFF
--- a/apps/chronicles/mad.config.ts
+++ b/apps/chronicles/mad.config.ts
@@ -32,7 +32,7 @@ export const config: MadConfig = {
         whatever: "",
     },
     about: {
-        endpoints: ["https://api.statoil.com/app/mad/test/api/v1"],
+        endpoints: [],
         buildNumber: getBuildNumber(),
     },
 };
--- a/apps/chronicles/mad.config.ts
+++ b/apps/chronicles/mad.config.ts
@@ -1,6 +1,7 @@
 import { MadConfig } from "@equinor/mad-core";
 import Logo from "./assets/images/icon.png";
 import { ImageSourcePropType } from "react-native";
+import { getBuildNumber } from "./settings";
 
 export const config: MadConfig = {
     appVersion: "1.0.0",
@@ -31,7 +32,7 @@ export const config: MadConfig = {
         whatever: "",
     },
     about: {
-        endpoints: [],
-        buildNumber: "",
+        endpoints: ["https://api.statoil.com/app/mad/test/api/v1"],
+        buildNumber: getBuildNumber(),
     },
 };
--- a/packages/core/src/components/screens/AboutScreen.tsx
+++ b/packages/core/src/components/screens/AboutScreen.tsx
@@ -2,6 +2,7 @@ import React from "react";
 import { Cell, EDSStyleSheet, Typography, useStyles } from "@equinor/mad-components";
 import { ScrollView, View } from "react-native";
 import { useAbout, useAppVersion, useEnvironment } from "../../store/mad-config";
+import { getMadCommonBaseUrl } from "../../utils/madCommonUtils";
 
 export const AboutScreen = () => {
     const styles = useStyles(themeStyles)
@@ -9,6 +10,7 @@ export const AboutScreen = () => {
     const environmentName  = environment.charAt(0).toUpperCase() + environment.slice(1);
     const appVersion = useAppVersion();
     const about = useAbout();
+    const endpoints = [getMadCommonBaseUrl(environment)].concat(about?.endpoints ?? []);
     
     return (
         <ScrollView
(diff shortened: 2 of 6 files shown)


packages/core/src/components/screens/AboutScreen.tsx
```diff
@@ -1,6 +1,63 @@
...
-import { Typography } from "@equinor/mad-components";
+import { Cell, EDSStyleSheet, Typography, useStyles } from "@equinor/mad-components";
+import { ScrollView, View } from "react-native";
+import { useMadConfig } from "../../store/mad-config";
 
 export const AboutScreen = () => {
-    return <Typography>Not implemented yet</Typography>;
+    const config = useMadConfig();
+    const styles = useStyles(themeStyles)
+    const environment = config.environment.charAt(0).toUpperCase() + config.environment.slice(1);
```
tormodAase, line 9: Not strictly necessary but we have a `useEnvironment` hook you can use to select environment from madconfig, and `useAppVersion` for appVersion. You can also create a new hook `useAbout` for getting the about-config. Up to you if you prefer sticking to `useMadConfig` or want to use the other hooks, but in my opinion it leads to easier to understand code if we leave out everything we don't need.

apps/chronicles/mad.config.ts
```diff
@@ -31,7 +32,7 @@ export const config: MadConfig = {
         whatever: "",
     },
     about: {
-        endpoints: [],
-        buildNumber: "",
+        endpoints: ["https://api.statoil.com/app/mad/test/api/v1"],
+        buildNumber: getBuildNumber(),
```
tormodAase, line 36: `react-native-device-info` has a method `getBuildNumber()` which picks it up from the app itself. This would let us get the build number in the About Screen without sending it in the mad-config. The only downside is it doesn't give you a value for web. Let's have a discussion on which approach we prefer @HaakonSvane @dansvend @HavardNot

tormodAase (approved):

packages/core/src/components/screens/AboutScreen.tsx
```diff
@@ -1,6 +1,63 @@
...
+                </Cell>
+                <Cell>
+                    <View style={styles.cellContainer}>
+                        <Typography variant="h2">Api</Typography>
+                        <View style={styles.columnContainer}>
+                            <Typography>Endpoints</Typography>
+                            <View>
+                                {config.about?.endpoints.map(endpoint => (
+                                    <Typography key={endpoint}>{endpoint}</Typography>
+                                ))}
```
tormodAase, line 40: One more thing. The app doesn't necessarily know about the mad api. I think it should still be displayed here, however. You can get it from `getMadCommonBaseUrl()`. I suggest removing it in `Chronicles` config too, to avoid duplicate entries

tormodAase (changes requested):

packages/core/src/components/screens/AboutScreen.tsx
dansvend: Thanks for the feedback 🙌 Switched to `useEnvironment` in [8773738](https://github.com/equinor/mad/pull/310/commits/87737389f1a61e1d48c5a6f543f1ebdf0fbbbc8a).

dansvend: Switched to `useAppVersion` in [d9bae07](https://github.com/equinor/mad/pull/310/commits/d9bae079a51f52a7a041ab3d0f4b9f8126f5db21).

dansvend: Switched to `useAbout` in [a8f98db](https://github.com/equinor/mad/pull/310/commits/a8f98db2e29a4742b4f173eb0e50361b699160a6).

packages/core/src/components/screens/AboutScreen.tsx
dansvend: Good point 👌 I added `madCommonBaseUrl` as a default endpoint in [5dfa832](https://github.com/equinor/mad/pull/310/commits/5dfa83200627b50cd6967b76caaae94784c5476d).

apps/chronicles/mad.config.ts
dansvend: I think we should aim for a platform agnostic solution. The current implementation is also config file type agnostic making it possible to use `env` or `json` files.

tormodAase: Lets go for that solution!

tormodAase (approved):

Issue #299, opened 2023-11-30
Circular Progress spinner freezes after short duration [web]


Jonas-Omdal:
### Package / App

MAD Components

### Version

0.8.1

### What happened?

The Circular progress is not working well on web. The spinner stops / freezes after a very short duration 🧊

**Expected Behaviour:**
The spinner should continuously rotate until the process is complete.

**Actual Behavior:**
The spinner stops and freezes after it spins one round.

**Impact:**
This freezing may lead users to believe the process has halted, potentially causing confusion.

### Other info

Check that the circular progress works well on both web and device.

PR #311, opened 2023-12-14
fix(circularprogess): adjust spinner for web

Jonas-Omdal:
Adjusted the circular progress animation so that it works properly on both web and device 

Fixes: #299


tormodAase (approved):
🔥 🔥 🔥

PR #312, opened 2023-12-14
feat: service message

tormodAase:
There are some tiny differences from the design:
- The "Remove" icon is an 'X'. I couldn't find an icon that resembled the icon in the design
- I added swipe to remove instead of pressing "Dismiss" to remove. I did this because I was scared the URL and "Dismiss" were too close to each other, and it could therefore be hard for the user to press the correct object.

Let me know if there's anything I should change @EmilieOlsen 

image
image
image
image

Placement on screens without headers (e.g. Login screen):
image


packages/core/src/components/service-message/ServiceMessageBanner.tsx
HaakonSvane, line 1: These files should be formatted. They are very hard to read. I see that the default workspace formatter was recently changed to eslint for some reason recently. I changed it back to prettier in this PR: https://github.com/equinor/mad/pull/317

maybe run this on all files to ensure they are properly formatted?

packages/core/src/components/service-message/ServiceMessageProvider.tsx
```diff
@@ -0,0 +1,23 @@
...
+    isDismissed: false,
+    setIsDismissed: () => undefined,
+    expansionEnabled: false,
+    setExpansionEnabled: () => undefined,
+    isExpanded: false,
+    setIsExpanded: () => undefined,
+    isError: false
+});
+
+export const ServiceMessageProvider = ({children}:PropsWithChildren) => {
```
HaakonSvane, line 15: Any specific reason to use a provider here? If we only render the banner in the custom navigation header, do we need to have this added complexity? Seems most of it can be moved directly into the banner component as well.

packages/core/src/components/service-message/components/ExpansionToggle.tsx
```diff
@@ -0,0 +1,17 @@
...
+export const ExpansionToggle = ({isExpanded}: ExpansionToggleProps) => {
+    const styles = useStyles(theme);
+    const icon = isExpanded ? "menu-up" : "menu-down";
+    return <Icon name={icon} style={styles.icon} size={24}/>
+}
+
+const theme = EDSStyleSheet.create((themeStyles) => ({
+    icon: {
+        color: themeStyles.colors.text.feedbackWarning,
+        marginRight: -16,
```
HaakonSvane, line 14: What is the reason for this negative margin?

packages/core/src/components/service-message/utils/resolveMessageFromServiceMessage.ts
```diff
@@ -0,0 +1,12 @@
+import { ServiceMessage } from "../types";
+
+export const resolveMessageFromServiceMessage = (serviceMessage: ServiceMessage, isError: boolean): { message: string, url?: string | null } => {
+    if (isError) return {
+        message: "Could not retrieve service message.",
```
HaakonSvane, line 5: Do we want a service message informing the user that there was an error fetching the service message? Should this not be added as a toast in the fetcher instead?

HaakonSvane (changes requested):
Really cool 😎 
Check my comments, especially regarding the formatting. It's a little hard to read code in github especially with the current formatting.

packages/components/src/styling/colors.ts
```diff
@@ -65,6 +65,7 @@ export const colors = {
     ui_background_dark_default: "#132634",
     ui_background_dark_darken: "#00000047",
     ui_background_dark_light: "#1D2731",
+    ui_background_dark_warning: "#FFC67A19",
```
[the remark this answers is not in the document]
tormodAase: It did look a bit weird in dark mode

packages/core/src/components/service-message/ServiceMessageProvider.tsx
tormodAase: We need some kind of global state management here. Otherwise the state will be reset every time you go to a new screen

packages/core/src/components/service-message/utils/resolveMessageFromServiceMessage.ts
tormodAase: I agree, however we do not have a toast system yet. I think for now this solution is acceptable, but we can create an issue on displaying a toast in the future?

packages/components/src/styling/colors.ts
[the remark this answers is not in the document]
EmilieOlsen: Agree, we should use the new EDS Color tokens. They are using a new system, so spending some time trying to figure it out 😅

packages/core/src/components/service-message/ServiceMessageProvider.tsx
HaakonSvane: Wouldn't this only happen if the component remounts on every navigation? Is this how it is working today?

packages/core/src/components/service-message/utils/resolveMessageFromServiceMessage.ts
HaakonSvane: Sure thing 🆗

packages/core/src/components/service-message/ServiceMessageProvider.tsx
tormodAase: The component will remount. We add a service message banner to all screens by default, but they are separate entities

packages/core/src/components/service-message/components/ExpansionToggle.tsx
tormodAase: I tried to correct for the padding from the container. However, the padding was incorrect. I switched the container padding to `dialog` instead, and removed this margin.

packages/core/src/components/service-message/utils/resolveMessageFromServiceMessage.ts
tormodAase: Issue created https://github.com/equinor/mad/issues/320

packages/core/src/components/service-message/ServiceMessageBanner.tsx
tormodAase: Ran formatting on most files

packages/components/src/styling/colors.ts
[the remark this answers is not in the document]
tormodAase: Old:
image

New (#7F4600):
image

packages/core/src/components/service-message/ServiceMessageProvider.tsx
HaakonSvane: Ahh. Thought the banner was stayed mounted during screen navigation

packages/core/src/components/service-message/ServiceMessageBanner.tsx
```diff
@@ -0,0 +1,103 @@
...
+                                <Typography
+                                    group="interactive"
+                                    variant="link"
+                                    style={styles.text}
+                                    onPress={() => void Linking.openURL(url)}
+                                >
+                                    {url}
+                                </Typography>
+                            )}
+                            <Typography
```
HaakonSvane, line 76: The previous text here read "click to remove" or something like that. Now that the text is no longer a link, should we remove the underline / change the styling since it is no longer interactive?

HaakonSvane (approved):
Good stuff pat'ner 🐄 🤠

packages/core/src/components/service-message/ServiceMessageBanner.tsx
tormodAase: @EmilieOlsen What do you think?

tormodAase: (Regarding the "Swipe to remove" text)

EmilieOlsen: Agree, remove the underline. But keep the boldness. In life.

HaakonSvane (approved):
Goodest shit 🎉

PR #317, opened 2023-12-17
feat(components): create `Chip` component

### LINKED ISSUE #315: Create `Chip` component
### Package / App

MAD Components

### Description

Should have the following props:
`onClick?`: callback method invoked when the user taps the chip. If not defined, the chip should not respond to touch events.
`onDelete?`: callback method invoked when the user taps the right cross. If defined, the cross should show in the chip
`variant?`: `"error" | "active" | "default" `. Determines the visuals of the chip
`disabled?`: boolean value determining whether or not the 
`iconName?`: icon name to use as leading icon for the chip

### Alternatives

_No response_

### Other info

From EDS:
Screenshot 2023-12-15 at 15 33 44

HaakonSvane:
- Implement `Chip` component
- Add `ChipScreen` in chronicles
- re-add Prettier as default workspace linter

closes #315


HaakonSvane:
Screenshot 2023-12-17 at 15 27 11

packages/components/src/components/Chip/Chip.tsx
```diff
@@ -0,0 +1,143 @@
...
+     * Callback method invoked when the user presses the delete button on the right side of the chip. If not defined, the chip does not render the delete button.
+     */
+    onDelete?: () => void;
+    /**
+     * Optional icon to render on the left side of the chip.
+     */
+    iconName?: IconName;
+};
+
+export const ICON_SIZE = 16;
```
HavardNot, line 37: I dont see this export being used anywhere else in the monorepo. Any reason to export it?

HavardNot (dismissed):
Heck yea brother! cheers from the dentist's office 😎

tormodAase (approved):
Smooth AF

packages/components/src/components/Chip/Chip.tsx
```diff
@@ -0,0 +1,143 @@
...
+     */
+    variant?: "default" | "error" | "active";
+    /**
+     * Whether or not the chip should respond to touch events and render its disabled state.
+     */
+    disabled?: boolean;
+    /**
+     * Callback method invoked when the user presses the chip. If not defined, the chip is not pressable.
+     */
+    onPress?: () => void;
```
PtjeEquinor, line 26: Its more clear the way it is now, but the disable prop is not really necessary here as onPress={undefined} disables it.

PtjeEquinor: ah yeah I see, only the pressable is disabled

packages/components/src/components/Chip/Chip.tsx
HaakonSvane: Shiee, good point

HaakonSvane: fixed in d2d0035

tormodAase (approved):

Issue #314, opened 2023-12-15
Add multiple text areas to the navigation cell 


Jonas-Omdal:
### Package / App

MAD Components

### Description

The current navigation cell only support having one title and description area. This setup is limiting when we want to provide more detailed or varied info. Add support for having multiple text areas to the navigation cell. Image for referance:

<img width="961" alt="Skjermbilde 2023-12-15 kl  10 51 05" src="https://github.com/equinor/mad/assets/83707387/eee9c52d-7f09-4946-a24a-f3d848fe637d">

PR #319, opened 2023-12-18
feat(cell): allow navigation cell to have additional titles

Jonas-Omdal:
Added new prop to the navigation cell that allows the usage of additional title / text sections

Fixes: #314

### DIFF
--- a/apps/chronicles/screens/components/CellScreen.tsx
+++ b/apps/chronicles/screens/components/CellScreen.tsx
@@ -76,6 +76,17 @@ export const CellScreen = () => {
                     iconName="navigation-outline"
                     onPress={() => null}
                 />
+                <Cell.Navigation
+                    title="This is a navigation cell with additional titles"
+                    description="And this is still its description!"
+                    additionalTitles={[
+                        { title: "Additional Title One" },
+                        { title: "Additional Title Two" },
+                        { title: "Additional Title Three" },
+                    ]}
+                    iconName="navigation-outline"
+                    onPress={() => null}
+                />
                 <MyCustomCell title="And this is a custom cell" />
             </Cell.Group>
             <Spacer />
@@ -126,6 +137,7 @@ export const CellScreen = () => {
                 </Typography>
             </View>
             <Spacer />
+
             <Cell onPress={() => null}>
                 <Typography>This cell responds to touch!</Typography>
             </Cell>
--- a/packages/components/src/components/Cell/NavigationCell.tsx
+++ b/packages/components/src/components/Cell/NavigationCell.tsx
@@ -29,6 +29,9 @@ export type NavigationCellProps = {
      * Name of the icon to use as a left adornment.
      */
     iconName?: IconName;
+    /**
+     * Additional titles to be shown in the cell.
+     */
     additionalTitles?: {
         title: string;
     }[];
@@ -65,7 +68,7 @@ export const NavigationCell = ({
             {...cellProps}
         >
             <View style={styles.contentContainer}>
-                <View style={styles.singleCellContainer}>
+                <View style={styles.titleDescriptionContainer}>
                     <Typography
                         group="cell"
                         variant="title"
@@ -106,7 +109,7 @@ export const NavigationCell = ({
 NavigationCell.displayName = "Cell.Navigation";
 
 const themeStyles = EDSStyleSheet.create(theme => ({
-    singleCellContainer: {
+    titleDescriptionContainer: {
         justifyContent: "center",
         gap: theme.spacing.cell.content.titleDescriptionGap,
     },
--- a/packages/components/src/components/Cell/NavigationCell.tsx
+++ b/packages/components/src/components/Cell/NavigationCell.tsx
@@ -29,6 +29,9 @@ export type NavigationCellProps = {
      * Name of the icon to use as a left adornment.
      */
     iconName?: IconName;
+    additionalTitles?: {
+        title: string;
+    }[];
 } & Omit<CellProps, "leftAdornment" | "rightAdornment" | "onPress">;
 
 export const NavigationCell = ({
@@ -37,6 +40,7 @@ export const NavigationCell = ({
     disabled = false,
     description,
     iconName,
+    additionalTitles,
     ...cellProps
 }: NavigationCellProps) => {
     const styles = useStyles(themeStyles);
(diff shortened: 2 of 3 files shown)


apps/chronicles/screens/components/CellScreen.tsx
```diff
@@ -76,6 +76,17 @@ export const CellScreen = () => {
                     iconName="navigation-outline"
                     onPress={() => null}
                 />
+                <Cell.Navigation
```
HaakonSvane, line 79: Maybe this should go into the `NavigationCellScreen` instead since it is more related to this variant than the generic `Cell` component.

packages/components/src/components/Cell/NavigationCell.tsx
```diff
@@ -61,24 +68,39 @@ export const NavigationCell = ({
...
+                            group="cell"
+                            variant="description"
+                            numberOfLines={2}
+                            color={disabled ? "textDisabled" : "textTertiary"}
+                        >
+                            {description}
+                        </Typography>
+                    )}
+                </View>
+                {title &&
```
HaakonSvane, line 91: On mobile, this looks a little strange:
Screenshot 2023-12-18 at 07 27 03

From the designs that i suspect that this change comes from, it looks like this on iPhone vs iPad:
Screenshot 2023-12-18 at 07 30 57


I suggest we try to style with these rules:

- On the main title line, the titles will be spread out as far away from each other as possible (i.e `justifyContent: "space-between"`
- If any title (normal or additional) comes into contact with any other, we move at least one additional title to the bottom of the cell

Screenshot 2023-12-18 at 07 41 06


Any thoughts on this @EmilieOlsen ?

packages/components/src/components/Cell/NavigationCell.tsx
```diff
@@ -29,6 +29,12 @@ export type NavigationCellProps = {
      * Name of the icon to use as a left adornment.
      */
     iconName?: IconName;
+    /**
+     * Additional titles to be shown in the cell.
+     */
+    additionalTitles?: {
+        title: string;
```
HaakonSvane, line 36: Since the prop name is `additionalTitles`, maybe we dont need each element in the list to be an object with a single `title` field. What do you think about just having a list of strings?

HaakonSvane (changes requested):
Great starting point 🥳 
I think we need to have a look at how we layout these additional titles on smaller devices. 
Maybe some input from design is also needed if we can't agree

packages/components/src/components/Cell/NavigationCell.tsx
EmilieOlsen: image

**For mobile:** 
The NormalTitle can be long, so there will probably not be any more room for AdditionalTitle, so this should go under the Description

**iPad:**
Another solution could be to use similar layout as mobile. AdditionalTitles under Description. Attached iPad layout is somewhat dependent on the length on the NormalTitle.

HaakonSvane: Last point is interesting, yes. It could be that the iPad title is long enough to cause issues as well. Either we add ellipsises to it in that case ("..."), or we have it automatically move the additional titles to the bottom as you mention.
If we want all cells to be consistent (i.e not have just one of them move the additional titles to the bottom), I think we should just cut off the main title to allow for the additional titles.

EmilieOlsen: Sounds good to me!

apps/chronicles/screens/components/components/NavigationCellScreen.tsx
```diff
@@ -47,6 +47,43 @@ export const NavigationCellScreen = () => {
...
+                title="This is a navigation cell with additional titles."
+                description="And this is still its description!"
+                additionalTitlesUnder={[
+                    "Additional Title One",
+                    "Additional Title Two",
+                    "Additional Title Three",
+                    "Additional Title Three",
+
+                    "Additional Title Three",
+                    "Additional Title Three",
```
HaakonSvane, line 81: Nitpicking, but should there seems to be some duplicated title texts here. Maybe just increment the numbers instead of duplicating the titles so that we don't confuse anyone with the component not rendering the third properly?

packages/components/src/components/Cell/NavigationCell.tsx
```diff
@@ -29,30 +28,64 @@ export type NavigationCellProps = {
...
         </View>
     );
 
+    const renderAdditionalTitles = (titles: string[]) => {
+        return titles?.map((text, index) => (
+            <Typography
+                onLayout={event => {
+                    const width = event.nativeEvent.layout.width;
+                    if (width > maxWidth) {
+                        setMaxWidth(width);
```
HaakonSvane, line 70: Setting the state of the component on each render like this has two consequences:
1. The component rerenders a lot of times depending on how many additional titles you have
2. Since you are using the same state variable for both titles under and to the side of the cell, the `maxWidth` variable will represent the largest title of them all. This means that if you have a very short title on the left, but a long one under, the left title will be affected by the titles under.

Instead of using state and layout calculations on the titles, is it possible to instead divide the side titles up into boxes of the same length, and overflow them with a ellipsises (...) if they go beyond this? You could use flex: 1 on each of the titles to ensure they each take up the same amount. Since you want them to align with the left side of the cell, you can use justifyContent: "flex-end" on the parent for example.

packages/components/src/components/Cell/NavigationCell.tsx
```diff
@@ -29,30 +28,64 @@ export type NavigationCellProps = {
...
         </View>
     );
-
     const DisclosureAdornment = () => (
         <View style={styles.adornmentContainer}>
             <Icon name="chevron-right" color={disabled ? "textDisabled" : undefined} />
         </View>
     );
 
+    const renderAdditionalTitles = (titles: string[]) => {
```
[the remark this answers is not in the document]
Jonas-Omdal: Something like this perhaps?

image

packages/components/src/components/Cell/NavigationCell.tsx
Jonas-Omdal: Can't recall why i did this in the first place. When removing the maxWidth and related logic it seems to be working just fine..

packages/components/src/components/Cell/NavigationCell.tsx
```diff
@@ -29,55 +46,97 @@ export type NavigationCellProps = {
...
-    );
+    const renderAdditionalTitlesRight = () =>
+        additionalTitlesRight?.map((text, index) => (
+            <Typography
+                key={index}
+                group="cell"
+                variant="title"
+                numberOfLines={1}
+                color="textTertiary"
+                style={[additionalTitlesRight && { flex: 1 }]}
```
HaakonSvane, line 78: is this conditional on the styling needed? It seems `renderAdditionalTitlesRight` already is conditionally called for.

HaakonSvane (commented):
Good stuff good stuff! Made a comment regarding a conditional that you can remove, but other then that it looks really good

HaakonSvane (approved):
👍

PR #340, opened 2024-01-08
fix!: remove unnecessary props from `ToggleButton`

Sandtory:
- Remove  `multiple` and `onChange`  props from `ToggleButtonProps` in `ToggleButton`.
- Remove props `Toggle`, `valid` and `isSelected` from `ToggleButtonContextContents` in `ToggleButton`.
- Add `activeIndex` prop to `ToggleButtonProps`.
- Update component render to follow the new prop changes.
- Add types file and corresponding props to button folder.
- Update `Button`'s toggle condition.


packages/components/src/components/Button/types.ts
```diff
@@ -0,0 +1,12 @@
...
+export type ToggleButtonContextType =
+    | {
+          /**
+           * Boolean value indicating whether or not the contexed button is selected or not.
+           */
+          isSelected: boolean;
+      }
+    | undefined;
+
+export type ButtonGroupChildrenType = React.ReactElement<ButtonProps> | boolean | null | undefined;
```
dansvend, line 12: It seems like this type is similar to `TabsChildrenType` and there will probably be new identical types. I would like a generic type `ChildrenType` where you enter `ButtonProps` into. Also having `| boolean | null | undefined` can be replaced by `ReactNode`. I would also move this type to where it belongs `ButtonGroup.tsx`.

dansvend (dismissed):
Nice job, captain 👨‍✈️ `Button` and `ToggleButton` is a lot cleaner now 🧼 I left some questions and improvement proposals 🙌

HaakonSvane (approved):
Looks good. I'm behind the other review comments (especially regarding the changelog) but other than that it looks awesome. For inspiration on how to write breaking changes, see the changelog for this library 👍

HaakonSvane (approved):
👍

Issue #364, opened 2024-01-23
[Styling] Cell.Group have unwanted paddingBottom when title is not in use


Jonas-Omdal:
### Package / App

MAD Components

### Version

0.11.2

### What happened?

Cell.Group is displaying unwanted bottom padding that should only be applied when the title prop is being used.

**Steps to Reproduce:**

1. Render the Cell.Group component without passing a title prop.
2. Observe the layout of the rendered component.

**Expected Behavior:**
When the Cell.Group component is rendered without a title, it should not have any bottom padding of the title container. 

![image](https://github.com/equinor/mad/assets/83707387/daa42a64-584a-4a15-849d-e350d240046d)

PR #366, opened 2024-01-23
fix(cellgroup): remove bottompadding when title is not in use

Jonas-Omdal:
Removed unwanted bottom padding when title is not in use in Cell.Group

Fixes: #364

### DIFF
--- a/.changeset/long-camels-compete.md
+++ b/.changeset/long-camels-compete.md
@@ -1,0 +1,5 @@
+---
+"@equinor/mad-components": patch
+---
+
+remove unwanted bottom padding when the title prop is not in use
--- a/.changeset/long-camels-compete.md
+++ b/.changeset/long-camels-compete.md
@@ -2,4 +2,4 @@
 "@equinor/mad-components": patch
 ---
 
-remove unwanted bottom padding when the title prop is not in use
+Updated Cell.Group and removed unwanted bottom padding when the title prop is not in use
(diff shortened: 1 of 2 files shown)


.changeset/long-camels-compete.md
```diff
@@ -0,0 +1,5 @@
+---
+"@equinor/mad-components": patch
+---
+
+remove unwanted bottom padding when the title prop is not in use
```
HaakonSvane, line 5: Maybe specify on what component is done on here

packages/components/src/components/Cell/CellGroup.tsx
```diff
@@ -62,10 +62,10 @@ export const CellGroup = ({
     );
 };
 
-const themeStyles = EDSStyleSheet.create(theme => ({
+const themeStyles = EDSStyleSheet.create((theme, props: CellGroupProps) => ({
```
HaakonSvane, line 65: I think it would be wise to create a separate `CellGroupStyleProps` that you can control separately from the `CellGroupProps`. If we were to add a new required prop to the `CellGroupProps`, it would now automatically require this new prop in the style object.

HaakonSvane (changes requested):
Awesome 🔥 
Some minor stuff to look at, but looks nice 🆗 Good catch btw

packages/components/src/components/Cell/CellGroup.tsx
Jonas-Omdal: Updated CellGroup now. Was it something like this you had in mind ?

HaakonSvane (approved):
Perfect 🙆‍♂️

HaakonSvane (approved):
👍

Issue #275, opened 2023-11-21
Upload page / Progress page


EmilieOlsen:
### Package / App

MAD Components

### Description

Showing the progress of creating/uploading items. If an error occurs, the user can expand the specific item/task to see the technical details of why it failed. Also has the opportunity to 'Copy to clickboard' if they want to give further feedback on this. 
Instead of showing all the 'sub-tasks', the number of successfully sub-tasks is shown + progress indicator shown the percentage of the upload (see frame 1)

<img width="1056" alt="image" src="https://github.com/equinor/mad/assets/111871753/cae331ff-8621-40c7-a422-25ed1489ae7b">

Figma link: https://www.figma.com/file/DddCDZoFicduHUyfRCSUjG4k/Inspection?type=design&node-id=3771%3A71392&mode=design&t=OWjAkg7zJoihnURo-1

PR #373, opened 2024-01-26
feat(components): add progress component

Jonas-Omdal:
Create Progress Screen

**Key changes:**
- New screen, Progress screen.
- New components, Progress and Progress.Item
      - Progress works as a wrapper for Progress.Item.
      - Progress.Item can contain one singel task or multiple tasks. Tasks have different statuses based on their progress.
- Reverted some animations related to the Circular progress that made the determine mode to act strange. This will therefor make the animation not work on web again..

Fixes: #275


apps/chronicles/hooks/useProgressUpload.ts
```diff
@@ -0,0 +1,100 @@
...
+    const [isSimulating, setIsSimulating] = useState(false);
+
+    const updateTaskStatus = (
+        taskIndex: number,
+        status: ProgressStatus,
+        errorDetails?: ProgressTaskErrorDetails,
+        newTitle?: string,
+    ) => {
+        setTasks(tasks =>
+            tasks.map((task, index) =>
```
HaakonSvane, line 38: I am getting some ESLint errors here regarding access on a `any` type. Should probably be fixed

apps/chronicles/hooks/useProgressUpload.ts
```diff
@@ -0,0 +1,100 @@
...
+                    : task,
+            ),
+        );
+    };
+
+    const resetTasks = () => {
+        setTasks(
+            tasks.map((task, index) => ({
+                ...task,
+                title: `CatWithHat${index + 1}.jpg`,
```
HaakonSvane, line 50: Here as well

apps/chronicles/hooks/useProgressUpload.ts
HaakonSvane: Nevermind. Reset and it seems to be gone now

apps/chronicles/hooks/useProgressUpload.ts
HaakonSvane: Nevermind

apps/chronicles/screens/components/components/ProgressScreen.tsx
```diff
@@ -0,0 +1,157 @@
...
+            <Typography>Progress with multiple tasks: </Typography>
+            <Spacer amount="small" />
+            <Progress title="Upload cat images">
+                <Progress.Item
+                    title="Upload images of cats with hats"
+                    description="uploading cats with hats"
+                    tasks={tasks}
+                    showCopyTextButton
+                    showRetryButton
+                    onCopyTextButtonPress={handleCopyErrorMessage}
```
HaakonSvane, line 115: Should we maybe just show the copy button if the `onCopyTextButton` prop is provided? This means there will be less props for the user to use so that they don't have to add both props (one for showing, and one for the onPress).

Same goes for the retry button

apps/chronicles/screens/components/components/ProgressScreen.tsx
HaakonSvane, line 1: The API you have created for these components looks absolutely beautiful. Super simple and gives the users a lot of control.

packages/components/src/components/Progress/Progress.tsx
```diff
@@ -0,0 +1,24 @@
...
+export type ProgressProps = {
+    /**
+     * Optional title for the progress group. This title can be used to provide a heading or context for the set of progress items contained within.
+     */
+    title?: string;
+    /**
+     * Children elements of the Progress component, which should be one or more `ProgressItem` components. The `Progress` component acts as a container that groups these items together.
+     * This allows for structured display of multiple progress-tracking elements, each representing a distinct task or process.
+     *
+     * The type `StrictChildrenReactNode<ProgressItemProps>` ensures that the children passed to `Progress` strictly adhere to the prop types defined by `ProgressItemProps`, promoting type safety and consistency within the component's usage.
```
HaakonSvane, line 15: Is this last comment regarding the typing of the children neccessary?

packages/components/src/components/Progress/ProgressExpandButton.tsx
```diff
@@ -0,0 +1,43 @@
...
+    taskCounter,
+    isExpanded,
+    toggleExpand,
+}: ProgressExpandButtonProps) => {
+    const styles = useStyles(themeStyles);
+
+    return (
+        <View>
+            {taskStatus !== "notStarted" && taskCounter > 0 && (
+                <Pressable style={[styles.dropDownContainer]} onPress={toggleExpand}>
```
HaakonSvane, line 27: Could this be a normal Button (from mad components) instead? We could use a ghost button for example. This way, we get more reuse of functionality and don't have to style or change this should we make changes to the visual expression of our other components in the future.

Also great for having the extra padding that comes with the buttons in regards to touch surface area.

packages/components/src/components/Progress/ProgressItem.tsx
```diff
@@ -0,0 +1,232 @@
...
+           * Manually set the overall status of the progress item. Valid statuses; 'notStarted', 'inProgress', 'success', or 'error'.
+           */
+          status: ProgressStatus;
+      };
+
+export type ProgressItemProps = {
+    /**
+     * Optional title for the progress item
+     */
+    title?: string;
```
HaakonSvane, line 42: Maybe nitpicking, but should we require the `title` prop instead? No title or description would mean an empty progress item, right?
Is having blank items needed / wanted? Maybe having it be required is good for ensuring users of the component stick to design standards

packages/components/src/components/Progress/ProgressItem.tsx
```diff
@@ -0,0 +1,232 @@
...
+
+    const handleRetryButtonPress = () => {
+        if (failedTask) {
+            onRetryButtonPress && onRetryButtonPress(failedTask);
+        }
+    };
+
+    const toggleExpand = () => {
+        setIsExpanded(!isExpanded);
+        LayoutAnimation.configureNext(animationConfig);
```
HaakonSvane, line 114: I can't seem to get the animation to work on web. Is this a known issue?
You could potentially look into using React Native Reanimated for animations as well. I know we probably should use that instead of the normal native one.
Don't stress this if it becomes too hard to implement though. Just a suggestion

packages/components/src/components/Progress/ProgressItem.tsx
```diff
@@ -0,0 +1,232 @@
...
+            </Typography>
+
+            {description ? (
+                <View style={styles.descriptionContainer}>
+                    {taskStatus !== "notStarted" && taskCounter > 0 ? (
+                        <Typography>
+                            {completedTaskCounter} / {taskCounter} {description}
+                        </Typography>
+                    ) : (
+                        <Typography>{description}</Typography>
```
HaakonSvane, line 145: Did the designs say something about the color of the description text if the task is not started? I see that the title is set to tertiary, but the description remains the same for these types of items.

packages/components/src/components/Progress/types.ts
```diff
@@ -0,0 +1,31 @@
...
+     */
+    title: string;
+    /**
+     * The current status of the task. This status should reflect where in the execution process the task currently is, such as 'notStarted', 'inProgress', 'success', or 'error'.
+     */
+    status: ProgressStatus;
+    /**
+     * Optional details about any error that has occurred during the task's execution. If provided, this should include a message detailing the error, and may also include an error code and suggestions for how to resolve the issue.
+     */
+    errorDetails?: ProgressTaskErrorDetails;
```
HaakonSvane, line 30: Maybe just rename this to `error` of type `ProgressTaskError`

packages/components/src/components/Progress/ProgressTaskItem.tsx
```diff
@@ -0,0 +1,66 @@
...
+export const ProgressTaskItem = ({ task, status }: ProgressTaskProps) => {
+    const styles = useStyles(themeStyles);
+    const token = useToken();
+
+    const taskInProgress = status === "inProgress";
+
+    const renderErrorDetails = ({ errorDetails }: ProgressTask) => {
+        return errorDetails ? (
+            <View style={styles.errorDetailsContainer}>
+                <Typography>{errorDetails.message}</Typography>
```
HaakonSvane, line 26: Is it intentional to render the error text as normal colored paragraph typography?
Visually it seems a little bit intense.

HaakonSvane: I am not saying we should fix it if this is what the designs require. If they say it should be like this, it is completely fine to merge this PR and fix it afterwards imo

packages/components/src/components/ProgressIndicator/useNoProgressAnimation.ts
HaakonSvane, line 1: Was the intention of this to fix the animation not looping on web? It seems to not be looping for me on web now as well.
As mentioned earlier, it might be easier to look into using React Native Reanimated for animations if you find yourself pulling your hair out trying to fix this issue with the normal react native

HaakonSvane (commented):
This is far beyond impressive Jonas. It simplifies as much as it can without compromising extensibility. Super great work. See my comments for some suggestions etc. Wow!

packages/components/src/components/Progress/ProgressItem.tsx
Jonas-Omdal: This should be fixed now. Changed to react-native-reanimated

packages/components/src/components/ProgressIndicator/useNoProgressAnimation.ts
Jonas-Omdal: I can look into that. Should this perhaps be addressed in a seperate issue ?

packages/components/src/components/Progress/ProgressExpandButton.tsx
Jonas-Omdal: Good idea

HaakonSvane (approved):
awesome 👍

packages/components/src/components/ProgressIndicator/useNoProgressAnimation.ts
HaakonSvane: Yep good idea 👍

PR #378, opened 2024-01-30
fix: add check to `WorkOrderCell` if value is truthy

Sandtory:
Add check if `PropertyRow` value is truthy, if it is not, dont render component.

### DIFF
--- a/.changeset/curly-doors-film.md
+++ b/.changeset/curly-doors-film.md
@@ -1,0 +1,5 @@
+---
+"@equinor/mad-dfw": patch
+---
+
+Add truhyness check to `PropertyRow` render function to avoid rendering rows with empty values.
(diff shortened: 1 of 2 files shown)


Jonas-Omdal (approved):
Looks good 😎👉👉

packages/dfw/src/dfwcomponents/WorkOrderCell/WorkOrderCell.tsx
```diff
@@ -148,19 +148,22 @@ export const WorkOrderCell = ({
                     ))}
                 <Label label={maintenanceType} style={{ marginBottom: 8 }} />
                 {Object.entries(rest).map(([key, value], index) => {
-                    const label = WorkOrderLabelMap[key as keyof WorkOrder];
-                    let displayValue = value;
-                    if (key === "basicStartDate" || key === "basicEndDate") {
-                        displayValue = value ? new Date(value).toLocaleDateString() : "";
+                    if (value) {
```
dansvend, line 151: I prefer to exit early by using `if (!value) return null;` to avoid a unnecessary nesting.

.changeset/curly-doors-film.md
```diff
@@ -0,0 +1,5 @@
+---
+"@equinor/mad-dfw": patch
+---
+
+Add truhyness check to `PropertyRow` render function to avoid rendering rows with empty values.
```
dansvend, line 5: Try to explain the problem instead of the implementation details.

PR #379, opened 2024-01-31
fix: menu doesnt rerender when children change

Sandtory:
Add update function from Floating UI to a useEffect that listens to children of `Menu`.


Jonas-Omdal (commented):
🔥

Jonas-Omdal (approved):

PR #382, opened 2024-02-01
fix: autocomplete crashes when multiselect unfocuses

Sandtory:
Fix `Autocomplete` to call for `internalTransform` before setting option to string.
Fix `MultiSelect` to handle `onBlur` differently on web so that menu persists when choosing an option.


packages/components/src/components/Autocomplete/Autocomplete.tsx
HaakonSvane, line 1: What caused the crash? I remember you saying something about an internal transform not being invoked, but is this something different ?

Sandtory: This is exactly that, the object that is set to the inputfield value couldnt accept the .toString function that was used on it originally. Now we use the internaltransform so that we can ensure that the object that is passed to selectedValue is properly converted to a string 🥸

HaakonSvane (approved):
👍

PR #388, opened 2024-02-01
fix(search): adjust animation on cancel button

Jonas-Omdal:
Adjusted the animation on `Search` so that it looks correctly when the cancel button is appearing.
Removed unused styling.


Sandtory (approved):
masterfully coded 🧑🏼‍🎓

PR #399, opened 2024-02-02
fix(search): prevent search to collapse

Jonas-Omdal:
Adjusted styling on Search to prevent that it collapses in some cases


HaakonSvane (approved):
(9️⃣ - ne) + (🧊 - i)

PR #407, opened 2024-02-08
fix: remove update `useEffect` from `Menu` component

### LINKED ISSUE #404: App crashes when using the Menu component on mobile device
### Package / App

MAD Components

### Version

0.12.5

### What happened?

The app crashes when using the Menu component.

**Steps to reproduce: **
1. Navigate to the Menu component.
2. Press the menu button and select any values from the menu.
3. Item is selected and you get a warning: measure cannot find view with tag #7793.
4. Press the menu again.
5. App crashes with following error: 
[1923, "RCTView",  {"left":"<<NaN>>", "top":263.5}] is not usable as a native method argument

Think this is caused by the useEffect in Menu.tsx.

### Other info

_No response_

### Relevant log output

_No response_

Sandtory:
closes #404

### DIFF
--- a/packages/components/src/components/Menu/Menu.tsx
+++ b/packages/components/src/components/Menu/Menu.tsx
@@ -47,7 +47,7 @@ export const Menu = ({
     children,
     ...rest
 }: React.PropsWithChildren<MenuProps & ViewProps>) => {
-    const { refs, floatingStyles, update } = useFloating({
+    const { refs, floatingStyles } = useFloating({
         sameScrollView: false,
         elements: {
             reference: anchorEl.current,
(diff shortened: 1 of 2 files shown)


packages/components/src/components/Menu/Menu.tsx
```diff
@@ -55,9 +55,6 @@ export const Menu = ({
         middleware: [offset(8), flip(), shift({ padding: 8 })],
         placement,
     });
-    useEffect(() => {
-        update();
```
Jonas-Omdal, line 59: By removing the useEffect is it necessary to define update here?
const { refs, floatingStyles, update } = useFloating({...

Jonas-Omdal (approved):
Noice

packages/components/src/components/Menu/Menu.tsx
Sandtory: true, i removed it now 🛠️

Jonas-Omdal (approved):

PR #411, opened 2024-02-09
feat: `mad-insights` now lets developers run track before app insights is initialized

tormodAase:


packages/insights/src/appInsights.ts
```diff
@@ -9,14 +9,16 @@ import { createBrowserHistory } from "history";
 import { Platform } from "react-native";
 
 import { obfuscateUser } from "./encrypt";
-import { AppInsightsInitConfig, Envelope } from "./types";
+import { AppInsightsInitConfig, Envelope, TrackEventPayload } from "./types";
 
 let appInsightsMain: ApplicationInsights;
```
dansvend, line 14: I think we should add `| undefined` to the type for this variable.

packages/insights/src/appInsights.ts
```diff
@@ -9,14 +9,16 @@ import { createBrowserHistory } from "history";
...
 let appInsightsLongTermLog: ApplicationInsights;
```
dansvend, line 15: I think we should add `| undefined` to the type for this variable.

dansvend (changes requested):

dansvend (approved):

PR #413, opened 2024-02-09
feat: `mad-core` now adds username to tracking after authenticating

tormodAase:


HavardNot (commented):
is this GDPR approved? Do we have this in the Expo-Core implementation?

tormodAase:
@HavardNot We do the same in [mad-expo-core](https://github.com/equinor/mad-expo-core/blob/main/src/screens/LoginScreen.tsx)

`setUsername` adds real name to main log, and obfuscated name to long term log

HavardNot (approved):

PR #436, opened 2024-02-19
fix(cell): strech out the pressable surface in additional surface

Jonas-Omdal:
Streched out the pressable surface in additional surface in the cell component


Sandtory (approved):
works as intended 👍

PR #446, opened 2024-02-21
fix: disable onpress when swiping

PtjeEquinor:
Disabeling the cell onPress when swiping is happening


HaakonSvane (approved):
Hella fresh cowboy 🤠

PR #467, opened 2024-03-06
chore: switch from `pnpm` to `yarn`

tormodAase:
While `pnpm` is great, and fast, we often run into issues with it because [expo does not have first class support for it](https://docs.expo.dev/guides/monorepos/). By switching to `yarn`, we can expect a smoother experience going forward.

`yarn install` is slower than `pnpm install`, but offers a more stable environment for us


tormodAase:
When/If this is approved, we probably have to force merge this one. I bet we wont be able to get the "Validate PR / 👁️ Validate code (pull_request_target)" to pass

HavardNot (dismissed):
Dont forget to create an ADR for this 😂

tormodAase:
ADR: https://github.com/equinor/mad-team/pull/391

HaakonSvane (dismissed):
Looks good. Tried changing components in dev mode against chronicles and it seemed to be working.
I left a comment in your ADR also.
Had a look into yarn workspaces docs and it seems they want us to reference internal package dependencies using `workspace:*`: https://yarnpkg.com/features/workspaces

HavardNot (approved):

Issue #343, opened 2024-01-09
Chip content not visible in dark mode in "Active" chips


tormodAase:
### Package / App

MAD Components

### Version

0.8.6

### What happened?

Light mode
![image](https://github.com/equinor/mad/assets/22524448/09c8f163-bc18-4e43-8792-3f792a526817)

Dark mode
![image](https://github.com/equinor/mad/assets/22524448/fce665ce-ba1c-4bb1-8b84-763df708378a)

Sandtory:
i think this is an issue for other components aswell where the background and text color used in darkmode is the same or very similar. 
Found it in `Menu`, `Autocomplete` and `Switch`.

PR #474, opened 2024-03-27
fix: correct color in dark mode for chip and menu

Jonas-Omdal:
Active chip had a issue where the background color and text color was the same.
Correct the text and background color in dark mode for chip and menu. 

**Key changes:**
- Added new color, `interactive_primary_dark_selected_highlight`
- Updated the color for text - menu - active - dark. 

It now looks like this:
image

Fixes: #343


HaakonSvane (approved):
Looks good 👍

PR #478, opened 2024-04-09
fix(components): export SelectMenu and SelectMenu.Multiselect  props

HaakonSvane:
- Add exports for prop types to SelectMenu and MultiselectMenu
- Add type exports to SelectMenu index file
- Add `Progress` component to docs
- Add `Button.Icon` component prop table to docs


Jonas-Omdal (approved):
Looks good 😎

PR #494, opened 2024-04-17
feat: poc expo authsession

HavardNot:
Created a proof-of-concept login button that authenticates users using Expo Auth Session.

Works for both mobile and web

This does not implement refreshtoken handling, authenticate silently, account management and other stuff. Will create new issues for these functionalities


tormodAase (changes requested):
Looks good, but I don't think we should merge this to main. That would be dangerous. We don't want to accidentally release a new version of auth/core with this code

HavardNot:
> Looks good, but I don't think we should merge this to main. That would be dangerous. We don't want to accidentally release a new version of auth/core with this code

Good point, i have removed all code changes that are outside of the auth package to make it as clean as possible for main. In the future we can add a config parameter to activate expo auth session in core and use that as a safe toggle for the authentictaion method.

tormodAase (approved):

tormodAase:
There's a conflict you need to resolve

HavardNot:
> There's a conflict you need to resolve

yes, also this PR needs to be merged before this one is ready for review
https://github.com/equinor/react-native-msal/pull/21

packages/auth/package.json
```diff
@@ -30,7 +30,7 @@
         "react": ">=18.0",
         "react-dom": ">=18.0",
         "react-native": ">=0.70",
-        "react-native-msal": "git+https://github.com/equinor/react-native-msal.git#d6b44b19454226d1bdc30a5d02d424d3e64a5163"
+        "react-native-msal": "*"
```
HaakonSvane, line 33: This will point to the react-native-msal package that is in npmjs i think (?).
Is this intended? I was under the impression that we need to use our own fork

packages/auth/package.json
```diff
@@ -33,7 +33,10 @@
         "react-native-msal": "*"
     },
     "dependencies": {
-        "@equinor/mad-components": "*"
+        "@equinor/mad-components": "*",
+        "expo-auth-session": "~5.0.2",
```
HaakonSvane, line 37: Is it enough to have expo auth session as a dependency and not a peer-dependency? If users will need to install this package in their own apps, we should perhaps move it to a peer dep

packages/auth/src/Expo-AuthSession/ExpoLoginButton.tsx
```diff
@@ -0,0 +1,91 @@
...
+            clientId,
+            scopes: scopes,
+            redirectUri,
+            responseType: ResponseType.Code,
+        },
+        discovery,
+    );
+
+    const onPress = () => {
+        promptAsync()
```
HaakonSvane, line 42: I would strongly advise using the async await paradigm for working with async methods. Code like this is very hard to read (nested `then`'s and `catch`es)

packages/auth/src/Expo-AuthSession/ExpoLoginButton.tsx
```diff
@@ -0,0 +1,91 @@
...
+import {
+    exchangeCodeAsync,
+    ResponseType,
+    useAuthRequest,
+    useAutoDiscovery,
+} from "expo-auth-session";
+import { MadAccount, UserInfo } from "../types";
+import { LoginButtonProps } from "../components";
+import { jwtDecode } from "jwt-decode";
+import "core-js/stable/atob";
```
HaakonSvane, line 13: Why is this import needed?

packages/auth/src/types.ts
```diff
@@ -8,3 +8,8 @@ export type MadAuthenticationResult = {
     account: MadAccount;
     accessToken: string;
 };
+export type UserInfo = {
```
HaakonSvane, line 11: Do we want to leak out this type to consumers of the package?
If this type is only needed to get typings on the parsing of the jwt-token, maybe it is better left closer to the decoding method.
I suggest maybe moving the `decodeToken` and this type to a separate file. This makes it much easier to write tests for as well (not that the decode method needs it unless we add more business logic to it)

HaakonSvane (dismissed):
👍 Looking forward to trying this out later on

packages/auth/package.json
HavardNot: this is an unchanged line that was temporarily changed to point to a specific commit to test for a fix that was needed for the react-native-msal package that was causing issues with the Code Inspection Github action. After the fix was tested and merged to main this dependency change was reverted.

If there are any changes that are required regarding this dependency I request @tormodAase to give input on this matter

packages/auth/package.json
HavardNot: These are indeed peer dependencies. moved them to peerdependency list

packages/auth/src/Expo-AuthSession/ExpoLoginButton.tsx
HavardNot: refactored function using await and surrounded in a single trycatch block

packages/auth/src/types.ts
HavardNot: created `decodeToken.ts` and moved UserInfo to the file as a private type

packages/auth/package.json
HaakonSvane: Oh, that explains why I saw the diff only for a certain commit.
As it stands right now (also before your PR then), it seems we are indeed pointing to the wrong instance of msal then, correct @tormodAase ?

HaakonSvane (dismissed):
Good stuff good stuff 🥇

packages/auth/package.json
HavardNot: It seems so, as the github action is failing once again due to the same error as before

packages/auth/src/Expo-AuthSession/useExpoAuthenticate.ts
HaakonSvane, line 1: Clean stuff 👌

HaakonSvane (approved):
Bestest stuff 💯

PR #495, opened 2024-04-17
feat(Cell): swipe buttons have access to swipe methods

tormodAase:
This PR also introduces e2e testing with [detox](https://wix.github.io/Detox/) in Chronicles!


apps/chronicles/components/ComponentNavigationCell.tsx
```diff
@@ -24,6 +24,7 @@ export const ComponentNavigationCell = ({
     return (
         <Cell.Navigation
             title={title}
+            testID={title}
```
HavardNot, line 27: Minor nitpick, should be renamed to `testId`

packages/components/src/components/_internal/SwipeableWithContext.tsx
```diff
@@ -0,0 +1,45 @@
...
+        throw new Error("Don't use legacy refs when referencing SwipeableWithContext");
+    const refToUse = ref ?? internalRef;
+    const swipeableMethods: SwipeableMethods = useMemo(
+        () => ({
+            close: logErrorIfUndefined(refToUse.current?.close),
+            openLeft: logErrorIfUndefined(refToUse.current?.openLeft),
+            openRight: logErrorIfUndefined(refToUse.current?.openRight),
+            reset: logErrorIfUndefined(refToUse.current?.reset),
+        }),
+        [refToUse, refToUse.current],
```
HavardNot, line 16: you can remove the dependency to refToUse.current as mutable values won't cause the component to re-render

apps/chronicles/components/ComponentNavigationCell.tsx
HavardNot: I guess this property name was not defined in this pr so you can ignore this comment.

HavardNot (changes requested):
looks like you got a warning from ESLint which should be rectified, other than that it looks good to me 🆒

apps/chronicles/components/ComponentNavigationCell.tsx
tormodAase: The property name is default in react, similar to `ref` and `key`

packages/components/src/components/_internal/SwipeableWithContext.tsx
tormodAase: Fixed

HavardNot (approved):

PR #504, opened 2024-04-23
feat: add text color support for `PropertyRow` and `WorkOrderCell`

### LINKED ISSUE #489: Add text color to `PropertyRow` value
### Package / App

MAD Components

### Description

We currently have a situation where text is needed to be a different color when conditions are met. With the current solution for PropertyRow, we cannot implement that.

Is it possible to get a text color for value? This can also be implemented in the workordercell under mad-dfw
Screenshot 2024-04-12 at 11 35 03

### Alternatives

_No response_

### Other info

_No response_

Sandtory:
Add TextColor prop to propertyrow
Add requiredEnd prop to workordercell
Add logic for conditionally rendering text color based on required end date

Breaking change:
Required end icon display now listens to RequiredEnd prop and not start & end date

### DIFF
--- a/.changeset/few-needles-tell.md
+++ b/.changeset/few-needles-tell.md
@@ -1,0 +1,5 @@
+---
+"@equinor/mad-dfw": minor
+---
+
+Add textcolor support for propertyrow and workordercell and add requiredEnd prop to workordercell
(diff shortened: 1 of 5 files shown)


Chibuzor-Nwemambu (approved):
Looks good 🚚

Issue #506, opened 2024-04-24
flex direction row do not work on Select


Jonas-Omdal:
### Package / App

MAD Components

### Version

0.14.2

### What happened?

The select component struggles when you use flex direction row.

![image](https://github.com/equinor/mad/assets/83707387/757c9356-5ed3-4a34-89cb-8c521c44a14b)

HaakonSvane:
I saw this as well during development. Managed to find a way around it in the inspection app. Let me know if you want to see how I fixed it there.

The issue actually also applies to the text inputs, since they share the same exact styling as the select component.

Jonas-Omdal:
> I saw this as well during development. Managed to find a way around it in the inspection app. Let me know if you want to see how I fixed it there.
> 
> The issue actually also applies to the text inputs, since they share the same exact styling as the select component. 

I fixed it for Select and Select.multi now. Only needed to add flex grow.

PR #507, opened 2024-04-24
fix(select): make select and multiselect responsive

Jonas-Omdal:
Prevent Select and MultiSelect to from "collapsing" when using flex direction row.

Fixes: #506


Chibuzor-Nwemambu (approved):
Nice! 🚚

PR #515, opened 2024-05-08
feat: expo auth session add persisting of token on web using msal-browser

HavardNot:


packages/auth/src/Expo-AuthSession/hooks/useAccount.ts
```diff
@@ -1,6 +1,29 @@
...
+import { getAccountWeb } from "../authenticationHandler";
+import { Platform } from "react-native";
 
-export const useAccount = () => {
+export const useAccountNative = () => {
     const authState = useAuth();
     return authState.userData;
 };
+
+export const useAccountWeb = () => {
```
tormodAase, line 12: Should these be exported?

tormodAase (approved):

packages/auth/src/Expo-AuthSession/hooks/useAccount.ts
HavardNot: probably not. no biggie tho

Issue #517, opened 2024-05-08
Wrong color on Seach text in dark mode


Jonas-Omdal:
### Package / App

MAD Components

### Version

0.14.3

### What happened?

Wrong color on Seach text in dark mode. In dark mode the text it black.

![image](https://github.com/equinor/mad/assets/83707387/247a9b2b-0e54-4c7c-8616-4946ad207289)

PR #519, opened 2024-05-08
fix(input): add color to input

Jonas-Omdal:
Corrected the text color in the`Input` component, which is being used in both `TextField` and `Search`. This should now work fine in both light and dark mode.

Fixes: #517


Sandtory (approved):
well written and documented code 🫡

PR #529, opened 2024-05-14
feat: deploy chronicles to radix

HavardNot:
Added deploy setup for hosting Chronicles Web in Radix:
https://web-chronicles-test.radix.equinor.com/
https://web-chronicles-prod.radix.equinor.com/

Currently using tags to push build requests to radix, feel free to add your own suggestions and opinions on how we should schedule releases to test and production 👍 

EDIT: Added release to production in the publish:all script and configured Radix to deploy to test from main


HaakonSvane:
I suggest following the repository-convention of releasing: pushing to prod when we merge the release PR.
This is also how the docs CI/CD are set up at the moment.
Would be something like
- Build to TEST whenever we merge to main
- Build to PROD whenever we merge the release PR


@tormodAase, what do you think?

package.json
```diff
@@ -43,7 +44,11 @@
         "format:all": "prettier --write .",
         "publish:all": "turbo run build && changeset version && changeset publish",
         "test:all": "turbo run test",
-        "clean:all": "turbo run clean && rm -rf node_modules/"
+        "clean:all": "turbo run clean && rm -rf node_modules/",
+        "push:tags": "git push origin --tags -f && echo Check status: https://console.radix.equinor.com/applications/chronicles/jobs",
```
HaakonSvane, line 48: Irrespective of how we decide to build for web, are these scripts needed in the package.json file? They seem to very pipeline-only related. Would there ever be a case where we'd need to use these locally? (push and fetch tags)

package.json
```diff
@@ -43,7 +44,11 @@
...
+        "fetch:tags": "git fetch origin --tags --force",
+        "release:test": "yarn fetch:tags && git tag -f TEST  && yarn push:tags",
```
HaakonSvane, line 50: Same goes for these I guess

HaakonSvane (approved):
💯 💯 
Looks fresh 👌 
I think we should discuss how we want to release as you mentioned. Made a comment in the PR for it.

package.json
HavardNot: These are used if you want to trigger a release manually. It is a way of deploying to Radix from whichever branch you are currently working on. DSD Mobile uses the same convention for all our web apps that are deployed to radix. 

Radix is normally configured so that each environment listens to a specific branch and deploys whenever anything gets pushed to the given branch. Without these commands you would have to manually merge the feature branch into the deploy branch every time you wish to deploy to a given environment. This is how they used to do it for the backend and it leads to lots overhead in handling merge conflicts and its easier to make mistakes. Today, backend uses a shell script to handle manual deployment to environments, which works mostly the same as these scripts.

You can read more about recommended best practices for CI in the Radix Documentation: https://www.radix.equinor.com/start/radix-concepts/#continuous-integration-and-deployment

It becomes a bit more complicated once we start implementing multiple apps in MAD Platform that deploy to web and are using Radix, but lets cross that bridge when we get there.

HavardNot: For reference, I implemented an alternative in the mad-docs repo that fully automates deployment as there are non-developers committing to the repo. https://github.com/equinor/mad-team/blob/main/radixconfig.yaml

For mad-docs, Radix deploys automatically to dev AUTOMATICALLY whenever someone pushes to a branch that uses the standard "issue-branch" naming convention: ("ISSUE_NUMBER"-"ISSUE_TITLE), and it deploys automatically to prod whenever a change is merged to main.

I dont think this is a good solution for mad or our other web apps as developers will constantly see their work get overwritten by other developers working in other branches (this still happens with solution above, but far less often, and developers are more aware when it happens as there is a manual step involved). Additionally, you might not always want to deploy for every commit, if its WIP f.ex.

tormodAase (dismissed):
LGTM!

package.json
tormodAase: I agree with @HaakonSvane that there are probably some ways we can improve releases, but I think it takes some work to set it up properly. I say we stick to manual release for this PR, and create an issue for improving releases

HaakonSvane (approved):
Looks hella nice 👌 🔥

Issue #521, opened 2024-05-08
Toast package


tormodAase:
### Package / App

Other

### Description

Requirements from discussion with @HavardNot , @dansvend , @HaakonSvane , @Jonas-Omdal a few months ago:

- Should be a separate package
- To make it easy for our apps to switch over, the API should be similar to [`react-native-salute`](https://github.com/equinor/react-native-salute/)
- `@equinor/mad-core` will consume the package and make it available to the apps
- Preferably stackable
- Preferably swipe to dismiss
- Use [`react-native-toast-message`](https://github.com/calintamas/react-native-toast-message/) if possible.

tormodAase:
`react-native-toast-message` looks fine, but it has some limitations:
- There's no support for multiple toasts at the same time (not sure if we wanted this).
- There's no queueing system, meaning if you try to display a toast when a toast is already visible, you will never display the 2nd toast.

I'll try to build a queueing system on top of it

PR #530, opened 2024-05-15
feat: toast package

tormodAase:


HavardNot (dismissed):
🍞

HavardNot (approved):

apps/chronicles/screens/ToastScreen.tsx
```diff
@@ -0,0 +1,96 @@
...
+        <ScrollView contentContainerStyle={styles.container}>
+            <View style={styles.readableContent}>
+                <Typography>
+                    Try out our different toast types by pressing the buttons below!
+                </Typography>
+                <View style={{ flexDirection: "row", gap: 16, marginTop: 8 }}>
+                    <Button
+                        title={"Success"}
+                        color="primary"
+                        onPress={() => addToast({ type: "SUCCESS", text: "Success!" })}
```
HaakonSvane, line 21: Was the API for this made so that the transition from the old package is easier?
We have to opportunity to require devs to use a new format (like lowercased types). I don't know if the convenience of only replacing the import and not changing the types is worth it. They have to change code in all their files anyways and TS should complain as a guide. Thoughts?

packages/components/src/styling/colors.ts
```diff
@@ -98,4 +98,23 @@ export const colors = {
     infographic_substitute_pink_rose: "#E24973",
     infographic_substitute_pink_salmon: "#FF92A8",
     infographic_substitute_purple_berry: "#8C1159",
+
+    //most of the toast colors are custom, not from EDS foundation
+    toast_error_background_light: "#FFD2D1",
```
HaakonSvane, line 103: What is the reason for these custom colors?

packages/toast/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2023 Equinor ASA
```
HaakonSvane, line 3: Should this be 2024 ?

packages/toast/src/components/ToastComponent.tsx
```diff
@@ -0,0 +1,50 @@
...
+    const { text: textColor } = getToastTokenForType(masterToken, type);
+
+    return (
+        <TouchableHighlight
+            disabled={typeof onPress !== "function"}
+            onPress={onPress}
+            style={{ borderRadius: 4 }}
+        >
+            <Paper elevation="raised" style={styles.background}>
+                <Icon name="information-outline" color={textColor} />
```
HaakonSvane, line 28: Should the icon either be configurable, or at least change based on the type of toast?
the information-outline for error, warning and success is maybe not so fitting

packages/toast/src/ToastEmitter.tsx
```diff
@@ -0,0 +1,20 @@
+import React from "react";
+import Toast, {
+    ToastProps as ImportedToastProps,
+    ToastConfig
+} from "react-native-toast-message";
+import { CustomToastTranslator } from "./components/CustomToastTranslator";
+import { ToastTypes } from "./types";
+
+const toastConfig: ToastConfig = {
+    [ToastTypes.ERROR]: params => <CustomToastTranslator {...params} />,
```
HaakonSvane, line 10: It seems like react-native-toast-message is the culprit behind this type of config. Is there no way around this? Seems redundant the way it is now since all configurations point to the same component

packages/toast/src/components/ToastComponent.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { getToastTokenForType } from "../utils/getToastTokenForType";
+
+export type ToastComponentProps = { text: string; type: ToastType; onPress?: () => void };
+export const ToastComponent = ({ text, type, onPress }: ToastComponentProps) => {
+    const styles = useStyles(theme, type);
+    const masterToken = useToken();
+    const { text: textColor } = getToastTokenForType(masterToken, type);
+
+    return (
+        <TouchableHighlight
```
HaakonSvane, line 22: What is the reason `TouchableHighlight` was used here instead of the mad-components `PressableHighlight`?
[react native docs](https://reactnative.dev/docs/touchablehighlight) state that the Pressable API is more future proof, which was the reason for why we made `PressableHighlight`. It also handles styling for the pressed state

HaakonSvane: Also, toasts that do not have the `onPress` method defined still behave as pressable components (darkens on touch). I think these toasts should not behave as pressable surfaces on user interaction

packages/toast/src/components/ToastComponent.tsx
HaakonSvane, line 1: Swiping behaviour on web is a little strange. I can't dismiss the toasts like on device for some reason.

packages/toast/src/utils/getMaxToastWidth.ts
HaakonSvane, line 1: The `Platform.select` method is a handy way of defining cases like this. Not that it is necesarrily worth it here, just mentioning it in case you haven't seen it.
Docs: https://reactnative.dev/docs/platform-specific-code#platform-module

packages/toast/src/utils/getToastTokenForType.ts
HaakonSvane, line 1: If you standardize the token types across the whole package (i.e moving away from the uppercase types as commented earlier), you can create much simpler resolvers like this in a single line by doing something like:
```
{
  backgroundColor: token.colors.toast[type].background
}
```

packages/core/README.md
HaakonSvane, line 1: This readme is becoming rather large. Would it be better to move this into sections in the mad docs and instead only include a minimal setup here?

packages/toast/README.md
HaakonSvane, line 1: Maybe we should move the more detailed documentation to the mad docs site?

packages/toast/package.json
```diff
@@ -0,0 +1,47 @@
+{
+    "name": "@equinor/mad-toast",
+    "version": "1.0.0",
```
HaakonSvane, line 3: Are you comfortable with major 1'ing this? Maybe we should keep it in 0 for some time to allow for more aggressive changes

HaakonSvane (changes requested):
Damn this is nice. Feels nice, looks nice, works nice 🔥
Made some comments here and there. Good job 👌

packages/components/src/styling/colors.ts
tormodAase: The design use these custom colors: https://www.figma.com/design/GmYi1qTwOkAM4ZyK5Bsj31/Component-library?m=dev&node-id=542%3A39794

packages/toast/src/components/ToastComponent.tsx
tormodAase: @EmilieOlsen What do you think?

packages/toast/src/utils/getMaxToastWidth.ts
tormodAase: Oh that's cool! Don't think it's necessary in this case since we are only changing one value

packages/toast/src/ToastEmitter.tsx
tormodAase: Don't think it's a way around it :(

packages/toast/LICENSE
tormodAase: Fixed

apps/chronicles/screens/ToastScreen.tsx
```diff
@@ -0,0 +1,96 @@
+import React from "react";
+import { Button, EDSStyleSheet, Typography, useStyles } from "@equinor/mad-components";
+import { ScrollView, View } from "react-native";
+import { addToast } from "@equinor/mad-toast";
+import { useCoreStackNavigation } from "@equinor/mad-core";
+
+export default function ToastScreen() {
```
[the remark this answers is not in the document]
tormodAase: Yup, fixed

apps/chronicles/screens/ToastScreen.tsx
tormodAase: I agree that we have the opportunity to change the format, but I don't like changing the format if there is no clear benefit of doing so. I think people will appreciate it if we keep the transition as easy as possible, [which is why we made it a goal to keep the api as similar as possible.](https://github.com/equinor/mad/issues/521) Maybe @dansvend @HavardNot @Jonas-Omdal has any thoughts on this

HaakonSvane: I understand. It is partially related to a comment I made below, but also comes from experience. Usually the enkleste er ofte det beste, and if there are two types of toasts floating around in the codebase (i.e. "ERROR" and "error"), then it quickly becomes confusing to work with.

packages/toast/src/components/ToastComponent.tsx
tormodAase: I'll switch it to `PressableHighlight`.

When it comes to disabling the pressable effect, do you have any suggestions for how to do so? It looks like `react-native-toast-message` gives the component a callback function for 'simplicity', but that makes it hard to check if we have a real callback. That is the reason why my "disabled" logic doesn't work

packages/toast/src/utils/getToastTokenForType.ts
tormodAase: Simplified `getToastTokenForType`

packages/toast/src/ToastEmitter.tsx
tormodAase: I simplified the config a bit now though, take a look

packages/toast/src/components/ToastComponent.tsx
tormodAase: Hmmm yeah something is off. I'll see if it's related to `react-native-toast-message`, or our code

packages/toast/package.json
tormodAase: Lowered the version number so we have space to squash some bugs. But I hope to major upgrade the package as soon as possible, since I believe this will be our most stable package

packages/core/README.md
tormodAase: Moved to mad-docs

packages/toast/src/components/ToastComponent.tsx
tormodAase: Fixed this by making the text unselectable on web. `userSelect: "none"`

apps/chronicles/screens/ToastScreen.tsx
tormodAase: After looking into `react-native-salute`, it looks like they in fact use lowercase, and our apps' `addToastByType` use uppercase. I'll change our toast types to lowercase

packages/toast/src/components/ToastComponent.tsx
EmilieOlsen: So sorry for the late reply! Didn't see this. 
I didn't think much of it until now and it makes sense. EDS has a separate error and a warning icon, so it makes sense to change it. 

image

apps/docs/docs/mad-toast/2-getting-started.md
```diff
@@ -0,0 +1,47 @@
...
+
+```tsx
+// Foo.jsx
+import { addToast } from "@equinor/mad-toast";
+import { Button } from "react-native";
+
+export function Foo(props) {
+    const showToast = () => {
+        addToast({
+            type: "success",
```
HaakonSvane, line 40: Will this type of prop passing work after you switched to enum-like typing?

apps/docs/docs/mad-toast/3-api.md
```diff
@@ -0,0 +1,34 @@
...
+| `onHide`         | Called when any Toast hides                                                                                                       | `() => void`      |               |
+| `onPress`        | Called on any Toast press                                                                                                         | `() => void`      |               |
+
+#### addToast
+
+The complete set of **options** is described below:
+
+| option     | description                                                          | type                         | required | default value |
+| ---------- | -------------------------------------------------------------------- | ---------------------------- | -------- | ------------- |
+| `type`     | Toast type. available values: `SUCCESS`, `ERROR`, `INFO`, `WARNING`. | `string`                     | yes      |               |
```
HaakonSvane, line 31: uppercases are not supported anymore i guess ?

HaakonSvane (approved):
Looks very nice. Awesome job on the docs migration 🎉 
Just some minor references in the toast docs that should be updated, but otherwise perfect 💯

HaakonSvane (approved):
✅

PR #534, opened 2024-05-22
feat: Add Mad OCR package

fnKode:
Adds the first version of the react-native-mad-tag-ocr package which enables tag scanning using the camera of a device. 

<strike>This PR also adds a new app called chrosetta for testing the tag scanner.</strike> 

The app can be tested in the mad-tag-ocr-testing-branch .The reason for creating a new app instead of using the existing chronicles app is due to some dependencies only being able to run in X86 simulators. Hence, by testing the new package in the chrosetta app we ensure that chronicles can still be tested using any simulator device. 
Thanks to @tormodAase for setting up the initial chrosetta app.


tormodAase (commented):
I'll leave this for now until we have an app to test the package in

fnKode:
Should a changeset be included for the first version of a package?

tormodAase (approved):
I don't know much about OCR, so I believe @HaakonSvane would be a better reviewer for this.

That said, looks good to me! Should probably remove the broken e2e tests in chrosetta

apps/chrosetta/e2e/cell.test.js
```diff
@@ -0,0 +1,33 @@
...
+    });
+
+    /*if you want some fun, uncomment the lines below*/
+    // it("SHOULD PLAY PING PONG ENDLESSLY", async () => {
+    //     while (true) {
+    //         await element(by.text("PING")).tap();
+    //         await element(by.text("PONG")).tap();
+    //     }
+    // });
+});
```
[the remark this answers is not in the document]
fnKode: Sounds good, I will remove the unnecessary code. Thanks for the review!

apps/chrosetta/radixconfig.yaml
HaakonSvane, line 1: It this radix file needed?

HaakonSvane: Guess it depends on whether or not we support OCR for web at the moment

packages/mad-tag-ocr/README.md
```diff
@@ -0,0 +1,53 @@
...
+```
+
+Note that the OCR camera only supports viewing in portrait mode.
+
+## ⚙️ Installation
+
+This library has the following peer dependencies. You have to install these in your project.
+
+```
+"@shopify/react-native-skia": "1.2.3",
```
HaakonSvane, line 43: Are all of these packages dealt with by Expo?
I know react native skia 1.2.3 is the version that Expo 51 recommends, so any lower Expo versions would complain if a user installed these versions in their project.
If these can be managed automatically by Expo, maybe just refer to the basic installation of each package instead of listing explicit version numbers to ensure we don't have to maintain these versions in the README all the time

packages/mad-tag-ocr/package.json
```diff
@@ -0,0 +1,55 @@
+{
+    "name": "@equinor/react-native-mad-tag-ocr",
+    "version": "1.0.0",
```
HaakonSvane, line 3: Do you feel comfortable majoring this to 1 on first release? If you suspect there will potentially be breaking changes to the package going forward, maybe majoring it to 0 to begin with is better.

packages/mad-tag-ocr/src/components/Camera/OCRCamera.tsx
```diff
@@ -0,0 +1,229 @@
...
+    formatTag,
+    getPainConfig,
+    getBoundingBox,
+    isPointInsideBoundingBox,
+    translatePointToFrame,
+} from "../../utils";
+import { SelectTagDialog } from "../SelectTagDialog";
+import { Button, EDSStyleSheet, useStyles } from "@equinor/mad-components";
+import { PopoverButton } from "../PopoverButton";
+import useStateToSharedValue from "../../hooks/useSharedState";
```
HaakonSvane, line 25: This should probably be a named export

packages/mad-tag-ocr/src/index.ts
```diff
@@ -0,0 +1 @@
+export { OCRCamera } from "./components/Camera/OCRCamera";
```
HaakonSvane, line 1: Maybe we should also export the ORCCameraProps to give users some flexibility for how they want to set up the camera

packages/mad-tag-ocr/src/utils.ts
```diff
@@ -0,0 +1,71 @@
...
+
+export const formatTag = (text: string) => {
+    "worklet";
+    return text
+        .split("")
+        .filter(char => new RegExp(/^[a-zæøå0-9-]/i).test(char))
+        .join("");
+};
+
+export function getPainConfig(color: Color) {
```
HaakonSvane, line 65: Should this be named `getPaintConfig` ?

packages/mad-tag-ocr/src/components/Camera/OCRCamera.tsx
HaakonSvane, line 1: Is it possible to split some of the code in this component into multiple files (subcomponents / hooks)?.

HaakonSvane (approved):
Very good job 💯 
One recommendation outside of the comments I left is that we expose the api (props that are exported) + docs to the docusaurus site like components and core has already.
Also, maybe create an initial changeset for the package.

apps/chrosetta/radixconfig.yaml
fnKode: Yes, I think we can remove that one since react-native-vision-camera currently only supports mobile device.

packages/mad-tag-ocr/README.md
fnKode: Seems like Skia is the only of these packages that are listed in the expo documentation https://docs.expo.dev/versions/latest/sdk/skia/

Good idea, I will update the readme

HaakonSvane (approved):
Good stuff 😎

packages/mad-tag-ocr/src/types.ts
```diff
@@ -9,6 +9,8 @@ export type Point = {
 export type BoundingBox = {
     topLeft: Point;
     bottomRight: Point;
+    width: number;
```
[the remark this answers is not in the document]
fnKode: > If it is for hiding the internal types that Skia expose to the user of the package, then we should, for consistency, create wrappers for the Color type as well. Might be more that I am missing, but that is at least what I can see so far.

Yes, it was done like that to hide the internal types.

I will look into making the types consistent

PR #538, opened 2024-05-24
chore: Update maintenance api wrapper to 1.30.0

fnKode:


tormodAase (approved):
LGTM!

But if you want a review from someone who knows what they are looking at, maybe @HaakonSvane should take a look

PR #550, opened 2024-06-11
feat(progress): add icon to progress tasks

Jonas-Omdal:
***Key changes:***
- Fixed issue where tasks with removed status was not included in the task counter in the `Progress`
component
- Added new `icon` and `iconColor` props, allowing users to display an icon on the right side of the

### DIFF
--- a/packages/components/src/components/Progress/ProgressItem.tsx
+++ b/packages/components/src/components/Progress/ProgressItem.tsx
@@ -62,8 +62,10 @@ export const ProgressItem = ({
 
     const styles = useStyles(themeStyles);
 
-    const taskCounter = tasks?.filter(task => task.status !== "removed").length;
-    const completedTaskCounter = tasks?.filter(task => task.status === "success").length;
+    const taskCounter = tasks?.length;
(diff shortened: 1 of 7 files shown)


packages/components/src/components/Progress/ProgressItem.tsx
```diff
@@ -62,8 +62,10 @@ export const ProgressItem = ({
 
     const styles = useStyles(themeStyles);
 
-    const taskCounter = tasks?.filter(task => task.status !== "removed").length;
-    const completedTaskCounter = tasks?.filter(task => task.status === "success").length;
+    const taskCounter = tasks?.length;
+    const completedTaskCounter = tasks?.filter(
+        task => task.status === "success" || task.status === "removed",
```
HaakonSvane, line 67: Maybe we should have a discussion with some designers about this, like we discussed.

HaakonSvane (approved):
👍

packages/components/src/components/Progress/ProgressItem.tsx
Jonas-Omdal: yes

PR #552, opened 2024-06-12
fix: Keyboard covering input fields in the feedback screen

fnKode:
If you select any input field in the feedback menu then you should now be able to scroll to the bottom of the page while also seeing the keyboard.

### DIFF
--- a/packages/core/src/components/screens/create-incident/CreateIncidentScreen.tsx
+++ b/packages/core/src/components/screens/create-incident/CreateIncidentScreen.tsx
@@ -72,7 +72,11 @@ export const CreateIncidentScreen = () => {
     };
 
     return (
-        <KeyboardAvoidingView behavior="padding" keyboardVerticalOffset={115}>
+        <KeyboardAvoidingView
+            behavior="padding"
+            keyboardVerticalOffset={115}
+            style={styles.container}
+        >
             <ScrollView contentInsetAdjustmentBehavior="automatic">
                 <Spacer />
                 <Cell>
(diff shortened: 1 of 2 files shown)


packages/core/src/components/screens/create-incident/CreateIncidentScreen.tsx
```diff
@@ -72,64 +72,66 @@ export const CreateIncidentScreen = () => {
...
-                    <View style={[styles.popupBox, styles.popupSuccess]}>
-                        <Typography>{`Ticket number: ${ticketNumber}`}</Typography>
-                    </View>
-                )}
-                {error && (
-                    <View style={[styles.popupBox, styles.popupDanger]}>
-                        <Typography>
-                            {/* eslint-disable-next-line @typescript-eslint/no-base-to-string,@typescript-eslint/restrict-template-expressions -- this rule sucks */}
-                            {`An error occurred creating your ticket: ${error}`}
+        <KeyboardAvoidingView behavior="padding" keyboardVerticalOffset={115}>
```
tormodAase, line 75: The `keyboardVerticalOffset` value seems a bit arbitrary. Does this solution work on all devices? iPhone 8, iPad mini, iPad Pro etc

tormodAase (dismissed):
Looks good! Take a look at my comment

packages/core/src/components/screens/create-incident/CreateIncidentScreen.tsx
fnKode: This is what we are using in notifications at least, and it seems to work fine for all the devices I have tried. But I will test a bit and verify that it works here as well 👍

fnKode: I tested some iphones and ipads and the styling was a bit weird, but I found a fix by adding `flex:1` to the `KeyboardAvoidingView`. Now it looks good on the two ipads and the two iphones 👍

tormodAase (approved):

PR #555, opened 2024-06-13
feat(components): add forward ref to ButtonCell component

HaakonSvane:
- Add forward ref to ButtonCell
- Write changeset


Jonas-Omdal (approved):
✨

Issue #561, opened 2024-06-19
`Cell`: dynamically adjusting the swipe groups does not work well on Web


Chibuzor-Nwemambu:
### Package / App

MAD Components

### Version

0.14.6

### What happened?

How to reproduce:

```tsx
const TestComponent = () => {
  const [checked, setChecked] = useState(false);

  return <Cell
                leftSwipeGroup={
                    checked
                        ? [
                              {
                                  color: "danger",
                                  title: "checked",
                                  onPress: methods => {
                                      setChecked(false);
                                      methods.close();
                                  },
                              },
                          ]
                        : []
                }
                rightSwipeGroup={
                    !checked
                        ? [
                              {
                                  color: "primary",
                                  title: "not checked",
                                  onPress: methods => {
                                      setChecked(true);
                                      methods.close();
                                  },
                              },
                          ]
                        : []
                }
            >
                <Typography>Hello world</Typography>
                <Button title="BUTTON"></Button>
            </Cell>
}
 ```

This works fine on iOS, but not on Web

https://github.com/equinor/mad/assets/75029767/5bdfa0d7-b91b-4aea-8480-3b685d373cd4

Swiping also sometimes triggers the cells onPress and leaves the cells color to be highlighted afterwards:

https://github.com/equinor/mad/assets/83707387/83c9760b-6131-4d1f-a925-341d49da28f5

PR #563, opened 2024-06-25
fix(cell): fix swipe related bugs

Jonas-Omdal:
***Key changes:***
- Fixed issue with on web where the cell failed to re-render after interaction.
- Fixed issue where swiping triggered the cells `onPress` event and left the cells color highlighted.
- Replaced the PressableHighlight animation with react-native-reanimated animations.

Fixes: #561

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -43,7 +43,8 @@
         "@testing-library/jest-native": "^5.4.3",
         "@testing-library/react": "^14.0.0",
         "@testing-library/react-native": "^12.3.0",
-        "react-test-renderer": "18.2.0"
+        "react-test-renderer": "18.2.0",
+        "react-native-gesture-handler": ">=2.0"
     },
     "files": [
         "CODE_OF_CONDUCT.md",
--- a/packages/components/src/components/Cell/Cell.tsx
+++ b/packages/components/src/components/Cell/Cell.tsx
@@ -1,7 +1,10 @@
-import React, { ReactNode, forwardRef, useContext, useState } from "react";
+import React, { ReactNode, forwardRef, useContext } from "react";
 import { View, ViewProps } from "react-native";
+import { TouchableWithoutFeedback } from "react-native-gesture-handler";
+import Animated from "react-native-reanimated";
 import { useStyles } from "../../hooks/useStyles";
 import { EDSStyleSheet } from "../../styling";
+import { useFadeAnimation } from "../../styling/animations";
 import { PressableHighlight } from "../PressableHighlight";
 import { SwipeableWithContext } from "../_internal/SwipeableWithContext";
 import { CellGroupContext, CellGroupContextType } from "./CellGroup";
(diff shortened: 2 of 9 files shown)


packages/components/src/styling/animations.ts
```diff
@@ -0,0 +1,40 @@
...
+    interpolateColor,
+} from "react-native-reanimated";
+import { useToken } from "../hooks/useToken";
+
+export const useFadeAnimation = () => {
+    const token = useToken();
+    const fadeAnimation = useSharedValue(0);
+
+    const handlePressIn = () => {
+        fadeAnimation.value = withTiming(1, {
```
HaakonSvane, line 14: I think the fading should only be applied on the pressOut. This seems to also animate the button on press in, which feels a little bit sluggish since there is a delay on every press

packages/components/src/components/Cell/Cell.tsx
```diff
@@ -66,8 +69,11 @@ export const Cell = forwardRef<View, React.PropsWithChildren<CellProps>>(
         ref,
     ) => {
         const { isFirstCell, isLastCell } = useContext(CellGroupContext);
+        const { handlePressIn, handlePressOut, animatedStyle } = useFadeAnimation();
+
         const styles = useStyles(themeStyle, { isFirstCell, isLastCell });
-        const [onPressDisabled, setOnPressDisabled] = useState(false);
+
+        const swipable = !!leftSwipeGroup || !!rightSwipeGroup;
```
HaakonSvane, line 76: i think it is written `swipeable`

HaakonSvane (changes requested):
Good stuff. Some minor comments around the code. Also found some issues that should be fixed when testing it around on iOS:

- Work order navigation cell in DFW package seems to not respond to tapping the cell

- All cells that have any touchable surface in an adornment (switch cell, custom cell with radio group in CellScreen etc) don't seem to respond to touch

- Animation on press in is (as commented as well) not desired since it makes the UI feel more sluggish

Jonas-Omdal:
> Good stuff. Some minor comments around the code. Also found some issues that should be fixed when testing it around on iOS:
> 
> * Work order navigation cell in DFW package seems to not respond to tapping the cell
> * All cells that have any touchable surface in an adornment (switch cell, custom cell with radio group in CellScreen etc) don't seem to respond to touch
> * Animation on press in is (as commented as well) not desired since it makes the UI feel more sluggish

Any clue why the adornments don't respond to any touches? I did some testing by replacing TouchableWithoutFeedback with Pressable, and it seemed to do the trick. However, this again caused swiping issues...

packages/components/src/components/Cell/Cell.tsx
```diff
@@ -189,4 +188,37 @@ const themeStyle = EDSStyleSheet.create((theme, props: CellGroupContextType) =>
         borderColor: theme.colors.border.medium,
         marginVertical: theme.spacing.menu.item.paddingVertical,
     },
+    children: {
+        flex: 1,
+        flexDirection: "row",
+    },
 }));
+
+/*
```
HaakonSvane: Commented code

packages/components/package.json
```diff
@@ -39,6 +39,7 @@
         "react-error-boundary": "^4.0.11"
     },
     "devDependencies": {
+        "jest-expo": "~49.0.0",
```
HaakonSvane, line 42: Can this be set to something less "chronicles-dependent" like "*"?
It will be a little bit hard to maintain this dependency whenever we bump the expo version of Chronicles

HaakonSvane (approved):
👍

packages/components/package.json
Jonas-Omdal: Tried changing it to "*", but then the test fails again:

image

PR #571, opened 2024-06-28
fix: Add selectable prop to PropertyRow

fnKode:
Adds a selectable prop to the PropertyRow component. This makes the value property selectable on iOS.


Sandtory (approved):
looks good!

PR #575, opened 2024-07-02
chore: update maintenance api wrapper

fnKode:
Updated to v1.31.0.

- Generated maintenance api code for v1.31.0
- Verified that it builds successfully
- Verified that no tests fail.


HaakonSvane (approved):
👍

Issue #585, opened 2024-07-29
Animations bugs after the expo-upgrade [react-native-reanimated]


Jonas-Omdal:
### Package / App

MAD Components

### Version

0.14.11

### What happened?

After upgrading react-native-reanimated from version 3.6.2 to 3.10.1, some of the animations from this package have started to act up, causing some unwanted behavior.

- [ ] Search component
Using the cancellable prop. 
1. Type something. 
2. Observe. The clear text button overlaps the cancel button.
![Simulator Screen Recording - iPhone 15 Pro Max - 2024-07-29 at 11 13 41](https://github.com/user-attachments/assets/ea8304c1-883d-4720-b02c-dd8ccd41832e)

- [ ] Progress component
1. Start simulation.
2. Expand the task list.
3. Observe. The list starts flickering. This stops when all of the tasks are completed. 
![Simulator Screen Recording - iPhone 15 Pro Max - 2024-07-29 at 11 22 51](https://github.com/user-attachments/assets/57444c77-2fa0-4fa5-8454-d70ade92e425)

PR #586, opened 2024-07-30
fix: fix search and progress animations

Jonas-Omdal:
**key changes**
- Updated animations for the `Search` and `Progress` components to prevent unwanted behaviour.

Fixes: #585


HaakonSvane (approved):
Real smooth 👍 
Tested and verified on iOS and web. Hella clean world class

Jonas-Omdal:
> Real smooth 👍 Tested and verified on iOS and web. Hella clean world class

Thanks  🙏🏼

PR #593, opened 2024-08-01
fix: corrected selection of expo auth session signout and react nativ…

HavardNot:
…e msal sign out

### DIFF
--- a/packages/core/src/components/screens/settings/useSignOut.ts
+++ b/packages/core/src/components/screens/settings/useSignOut.ts
@@ -1,17 +1,18 @@
 import { useCallback } from "react";
-import { useCoreStackNavigation } from "../../../hooks/useCoreStackNavigation";
+import { useCoreStackNavigation } from "../../../hooks";
 import { ExpoAuthSession, signOut } from "@equinor/mad-auth";
 import { alert } from "@equinor/mad-components";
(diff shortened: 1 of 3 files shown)


packages/core/src/components/screens/settings/useSignOut.ts
```diff
@@ -1,17 +1,18 @@
...
+import { useDemoMode, getConfig } from "../../../store";
 
 export const useSignOut = () => {
     const navigation = useCoreStackNavigation();
     const demoMode = useDemoMode();
     const signOutFn = useCallback(async () => {
         try {
-            const result = getConfig() ? ExpoAuthSession.signOut() : await signOut();
+            const result = getConfig().experimental?.useExpoAuthSession
+                ? ExpoAuthSession.signOut()
```
tormodAase, line 14: Should `ExpoAuthSession.signOut()` also be awaited?

tormodAase (approved):

packages/core/src/components/screens/settings/useSignOut.ts
HavardNot: no. most expoauthsession function equivalents do not need to be awaited. Double checked and the logout function of EAS is not async

PR #596, opened 2024-08-06
feat(skia): add image encoding options to makeImageSnapshot

### LINKED ISSUE #595: Add encoding options for `makeImageSnapshot`
### Package / App

React Native Skia Draw

### Description

Current snapshot handler only encodes to PNG. We should add support for configuring which encoding to use as some apps require JPG encoding.

### Alternatives

Encoding in the apps. Redundant and will increase list of dependencies. RN Skia encoding should probably be faster as well.

### Other info

_No response_

HaakonSvane:
- Add `encodingOptions` parameter to `makeImageSnapshot` method
- Add `data` field in `makeImageSnapshot` method that only contains base64 image data

closes #595

### DIFF
--- a/packages/skia-draw/src/CanvasControlProvider/types.ts
+++ b/packages/skia-draw/src/CanvasControlProvider/types.ts
@@ -1,6 +1,6 @@
 import type { Color, SkRect } from "@shopify/react-native-skia";
 import type { CanvasTool } from "../Canvas/types";
-import { SkiaDrawSnapshot } from "../types";
+import { ImageSnapshotEncodingOptions, SkiaDrawSnapshot } from "../types";
 
 export type CanvasControlProviderProps = {
     /**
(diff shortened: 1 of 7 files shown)


packages/skia-draw/src/CanvasControlProvider/types.ts
```diff
@@ -63,7 +63,10 @@ export type CanvasControlContextValues = {
 export type CanvasControls = {
     undo: () => void;
     clear: () => void;
-    makeImageSnapshot: (rect?: SkRect) => SkiaDrawSnapshot | undefined;
+    makeImageSnapshot: (
+        rect?: SkRect,
+        encodingOptions?: ImageSnapshotEncodingOptions,
+    ) => SkiaDrawSnapshot | undefined;
```
tormodAase, line 69: If the user wants to add `encodingOptions`, but not `rect`, they would have to do:

```ts
makeImageSnapshot(undefined, encodingOptions)
```

which is fine, but should we consider putting the arguments into an object instead?

It would be a breaking change, but I think it would be a better solution overall.

```ts
makeImageSnapshot({
  encodingOptions: someEncodingOptions,
  rect: someRect
})
```

tormodAase (dismissed):
Looks yummy 😋

packages/skia-draw/src/CanvasControlProvider/types.ts
HaakonSvane: Yeah, I did it like this to avoid breaking changes, but if you prefer changing the signature, I'll go ahead and change it

HaakonSvane: fixed in 627f9b7

tormodAase (approved):
🤤🤤🤤🤤🤤

PR #599, opened 2024-08-09
feat(progress)!: remove `removed` status on progress item and decendants

HaakonSvane:
- Fix issue where items that have tasks containing only `success` and `notStarted` were summarized as
- Write unit tests for `summarizeStatuses`
- Remove `removed` status


Jonas-Omdal (approved):
Nice 😎

PR #602, opened 2024-08-12
feat(maintenance-api): switch to fetch client

### LINKED ISSUE #600: Change from axios client to node fetch API in api wrapper
### Package / App

Maintenance API typescript wrapper

### Description

Current wrapper uses axios for fetching. We should move to using fetch instead as axios is no longer such a good alternative (current version comes with some security vulnerability for example)

### Alternatives

_No response_

### Other info

_No response_

### LINKED ISSUE #601: Ensure token is sent under automatic api redirects
### Package / App

Maintenance API typescript wrapper

### Description

Current implementation accepts redirects, but does not embed the token on the new request. We should send the token as well after talks to the maintenance api

### Alternatives

_No response_

### Other info

_No response_

HaakonSvane:
- move to using the fetch API client (was axios)
- remove old dependencies
- try fixing the redirect issue where authorization headers are not included under redirects

closes #601, closes #600

### DIFF
--- a/.changeset/cold-hats-repeat.md
+++ b/.changeset/cold-hats-repeat.md
@@ -1,0 +1,5 @@
+---
+"@equinor/mad-maintenance-api-ts-wrapper": patch
+---
+
+Moved to using the fetch API client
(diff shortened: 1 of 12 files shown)


Jonas-Omdal (approved):
Noice

PR #605, opened 2024-08-16
fix(toast): fix error where toasts could be locked for eternity

### LINKED ISSUE #604: Invoking `addToast` before `ToastEmitter` causes all future toasts to not show
### Package / App

Other

### Version

mad-toast@latest

### What happened?

If `ToastEmitter` is rendered after a call to `addToast`, all subsequent calls to `addToast` results in nothing happening. This bug was discovered first in the WorkOrders app after a toast is emitted from failed login-attempts that are done higher up in the render hierarcy

### Other info

This bug seems to occur from the locking/unlocking mechanism in the toast queue. If an `addToast` call is made without an emitter, the `onHide` callback of the toast is never invoked, meaning that the queue will be locked for ever.

### Relevant log output

_No response_

HaakonSvane:
- move to `useSyncExternalStore` for queue system

closes #604

### DIFF
--- a/packages/toast/src/queue.ts
+++ b/packages/toast/src/queue.ts
@@ -1,14 +1,30 @@
 import { AddToastOptions } from "./types";
 
+type Subscriber = () => void;
+
 export class ToastQueue {
     private static queue: AddToastOptions[] = [];
-    private static isLocked = false;
-
-    public static lock = () => (this.isLocked = true);
-    public static unlock = () => (this.isLocked = false);
-    public static push = (item: AddToastOptions) => this.queue.push(item);
+    private static subscribers: Set<Subscriber> = new Set<Subscriber>();
+    private static emitChange = () => {
+        for (const subscriber of this.subscribers) {
+            subscriber();
+        }
+    };
+    public static subscribe = (subscriber: Subscriber) => {
+        this.subscribers.add(subscriber);
+        return () => {
+            this.subscribers.delete(subscriber);
+        };
+    };
+    public static getSnapshot = () => this.queue;
+    public static enqueue = (item: AddToastOptions) => {
+        this.queue = [...this.queue, item];
+        this.emitChange();
+    };
     public static shift = () => {
-        if (this.isLocked) return undefined;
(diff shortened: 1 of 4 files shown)


packages/toast/src/queue.ts
```diff
@@ -1,14 +1,30 @@
...
+    public static getSnapshot = () => this.queue;
+    public static enqueue = (item: AddToastOptions) => {
+        this.queue = [...this.queue, item];
+        this.emitChange();
+    };
     public static shift = () => {
-        if (this.isLocked) return undefined;
-        return this.queue.shift();
+        const toast = this.queue.at(0);
+        this.queue = [...this.queue.slice(1)];
```
tormodAase, line 26: Could this be simplified with `const toast = this.queue.shift();`?

packages/toast/src/queue.ts
```diff
@@ -1,14 +1,30 @@
...
+    };
+    public static subscribe = (subscriber: Subscriber) => {
+        this.subscribers.add(subscriber);
+        return () => {
+            this.subscribers.delete(subscriber);
+        };
+    };
+    public static getSnapshot = () => this.queue;
+    public static enqueue = (item: AddToastOptions) => {
+        this.queue = [...this.queue, item];
```
tormodAase, line 21: I guess we need completely new objects for `useSyncExternalStore` to work?

packages/toast/src/queue.ts
tormodAase: Or do we need a completely new array for this to work?

tormodAase (approved):
🍾

packages/toast/src/queue.ts
HaakonSvane: Yeah. Took me a while before I caught that initially.
They need to be fresh copies every time for the hook to detect store changes

packages/toast/src/queue.ts
HaakonSvane: Yup. As in my other comment, I had to figure this out the hard way

Issue #607, opened 2024-08-19
Update Cell component to use PressableHighlight when not swipeable


Sandtory:
### Package / App

MAD Components

### Version

Latest

### What happened?

The issue appears when using a mix of react native gesture handler and react native pressable components. An example of this would be in the Cell component where we are using TouchableWithoutFeedback in order to make the swipeable prop function properly. This creates problems further down the line, if you were to use the cell in a menu the rootmodal of the menu would overlay the TouchableWithoutFeedback of the cell and the user would not be bale to press it. 

- [ ]  Upgrade gesture handler package, the swipeable will have a new import path from react-native-gesture-handler/ReanimatedSwipeable (Dont go beyond expo support version, npx expo install "package")
- [ ] split cell implementation based on swipeable props, if swipeable use react native gesture handler, else pressablehighlight
- [ ] Test thoroughly and well, all components that are related to cells need to be tested

PR #608, opened 2024-08-21
fix(cell): use gesture handler when the cell i swipeable

Jonas-Omdal:
**Key changes:**
- Replaced `PressableHighlight` with `TouchableWithoutFeedback` when the cell is swipable
- Fixed bug where the divider in Cell.Group was cropped if you have a cell with `additionalSurface` prop.

Fixes: #607

### DIFF
--- a/packages/components/src/components/Cell/Cell.tsx
+++ b/packages/components/src/components/Cell/Cell.tsx
@@ -75,7 +75,7 @@ export const Cell = forwardRef<View, React.PropsWithChildren<CellProps>>(
 
         const swipeable = !!leftSwipeGroup || !!rightSwipeGroup;
 
-        const CellContent = () => (
+        const cellContent = () => (
             <>
                 <View style={styles.contentContainer}>
                     {leftAdornment && <View style={styles.adornment}>{leftAdornment}</View>}
@@ -93,7 +93,7 @@ export const Cell = forwardRef<View, React.PropsWithChildren<CellProps>>(
             </>
         );
 
-        const Cell = () => (
+        const renderCell = () => (
             <View {...rest} style={[styles.container, rest.style]} ref={ref}>
                 <View style={{ flexDirection: "row" }}>
                     {additionalSurface && (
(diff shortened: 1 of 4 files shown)


Sandtory (dismissed):
great work! you are the mithocondria of this cell ⚛🦠

packages/components/src/components/Cell/Cell.tsx
```diff
@@ -75,6 +76,24 @@ export const Cell = forwardRef<View, React.PropsWithChildren<CellProps>>(
...
+                </View>
+                {!isLastCell && (
+                    <View style={styles.dividerOuter}>
+                        <View style={styles.dividerInner} />
+                    </View>
+                )}
+            </>
+        );
+
+        const Cell = () => (
```
HaakonSvane, line 96: Having a component inside the Cell component with the same name can be a little bit confusing. Maybe find a better suited name for it?

packages/components/src/components/Cell/Cell.tsx
```diff
@@ -138,10 +152,10 @@ export const Cell = forwardRef<View, React.PropsWithChildren<CellProps>>(
                     ))
                 }
             >
-                {CellContent()}
+                <Cell />
```
HaakonSvane, line 155: There is a functional difference between referring to sub-components with the jsx-style syntax (`<Something/>`) vs calling it as a method `{Something()}`. If you use jsx-style, the component will be treated by react as a separate component in the rendering tree, and you should make sure that the sub-component uses __no__ state from the parent. Calling it as a method is faster, but if the sub-component becomes large, it is better to move it completely out as a separate component instead.

The rule of thumb to use is:
If you want to treat the component as a standalone react component:
- Use JSX-style syntax
- Move the component completely outside of the parent and pass all required data as props

If your sub-component is small and you want to keep it inside the parent:
- Use method syntax
- Freely use variables defined in either the sub-component or in the parent

HaakonSvane (approved):
Looks good 💯 Glad to see this fixed. I'd have a quick look at my comments and see if you want to make any changes before merging

HaakonSvane (approved):
💯

Issue #613, opened 2024-09-04
Deviations on the WorkOrderCell component


Jonas-Omdal:
### Package / App

MAD Components

### Version

0.15.1

### What happened?

Deviations:
- In addition to the `start` and `complete` button there should also be a TECO (Technical complete) button.
- Functional location is missing
- Incorrect titles:
        - Every word in the in the titles starts with a capital letter. Only the first word should have this.
        - Basic Start Date --> Basic start
        - Basic End Date --> Basic end
                - It might be better to have these on one line: Basic start / finish
        - Work Center ID --> Main work center
- Unable to change / overwrite styles on the cell. It would be nice to set the border widths to 0

PR #614, opened 2024-09-08
fix(workordercell): improved component and  fixed deviations

Jonas-Omdal:
**Key changes:**
- Created new component `WorkOrderCell.Navigation`
- Updated showSymbols props: status icons can now be shown in both row and column
- Updated some of the text in the default labels
- Added new prop, overwriteLabel where you can overwrite the default labels
- Added possibility to overwrite the style on the Cell itself, for example the border width.

Fixes: #613

### DIFF
--- a/packages/dfw/src/dfwcomponents/WorkOrderCell/PropertyList.tsx
+++ b/packages/dfw/src/dfwcomponents/WorkOrderCell/PropertyList.tsx
@@ -1,0 +1,95 @@
+import React from "react";
+import { PropertyRow } from "../PropertyRow";
+import { WorkOrder } from "./types";
+import moment from "moment";
+import { Color, useToken } from "@equinor/mad-components";
+
+const defaultWorkOrderLabelMap: Record<keyof WorkOrder, string> = {
+    title: "Title",
+    workOrderId: "Work order ID",
+    maintenanceType: "Maintenance type",
+    tagId: "Tag ID",
+    equipmentId: "Equipment ID",
+    activeStatusIds: "Active status IDs",
+    basicStartDate: "Basic start",
+    basicEndDate: "Basic end",
+    requiredEnd: "Required end",
+    workCenterId: "Work center ID",
+    functionalLocation: "Functional Location",
+} as const;
+
+type PropertyListProps = {
+    workOrder: Partial<WorkOrder>;
+    valueColor: Color;
+    currentDate: Date;
+    overwriteLabel?: Partial<Record<keyof WorkOrder, string>>;
+};
+
+export const PropertyList = ({
+    workOrder,
+    valueColor,
+    currentDate,
+    overwriteLabel = {},
+}: PropertyListProps) => {
+    const requiredEnd = workOrder.requiredEnd ? new Date(workOrder.requiredEnd) : null;
+
+    const formatDate = (dateString: string) => moment(dateString).format("DD.MM.YYYY");
+    const token = useToken();
+
+    const getLabel = (key: keyof WorkOrder, combinedDates: string | null) => {
+        if (key === "basicStartDate" && combinedDates) {
+            return "Basic start / end";
+        }
+        return overwriteLabel?.[key] ?? defaultWorkOrderLabelMap[key];
+    };
+
+    const getDisplayValue = (
+        key: keyof WorkOrder,
+        value: string | undefined,
+        combinedDates: string | null,
+    ) => {
+        if (key === "basicStartDate" && combinedDates) {
+            return combinedDates;
+        }
+        if (["basicStartDate", "basicEndDate", "requiredEnd"].includes(key)) {
+            return value ? formatDate(value) : "";
+        }
+        return value;
+    };
+
+    const combinedDates =
+        workOrder.basicStartDate && workOrder.basicEndDate
+            ? `${formatDate(workOrder.basicStartDate)} - ${formatDate(workOrder.basicEndDate)}`
+            : null;
+
+    return (
+        <>
+            {Object.entries(workOrder).map(([key, value], index) => {
+                const typedKey = key as keyof WorkOrder;
+
+                if (combinedDates && typedKey === "basicEndDate") return null;
+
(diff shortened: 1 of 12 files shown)


.changeset/few-tigers-grow.md
```diff
@@ -0,0 +1,6 @@
+---
+"@equinor/mad-dfw": major
```
tormodAase, line 2: When working on packages with version 0.X.X, you can do a `minor` bump instead, once you enter v1, the package is considered `stable`

packages/dfw/src/dfwcomponents/WorkOrderCell/types.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export type WorkOrder = {
+    title: string;
+    workOrderId: string;
+    maintenanceType?: string;
+    tagId?: string;
+    equipmentId?: string;
+    activeStatusIds?: string;
+    basicStartDate?: string;
+    basicFinishDate?: string;
```
tormodAase, line 12: what does `basic` mean in this context?

Jonas-Omdal: ¯\_(ツ)_/¯
But, in the api and figma design it follow this naming.

packages/dfw/src/dfwcomponents/WorkOrderCell/WorkOrderCell.tsx
```diff
@@ -1,222 +1,128 @@
...
+                            disabled={isCompleteDisabled}
+                            onPress={onCompleteButtonPress}
+                        />
+                    )}
+                    {onTecoButtonPress && showActions.tecoButton && (
+                        <Button
+                            title="Technical complete"
+                            variant="outlined"
+                            onPress={onTecoButtonPress}
+                        />
```
tormodAase, line 89: Can all these buttons be available at the same time?

Also, since the buttons' visibility is dependent on `onStartButtonPress`, `onCompleteButtonPress`, and `onTecoButtonPress` respectively, is `showActions` necessary?

Jonas-Omdal: Yes, however, there are specific conditions that must be met for these buttons to be displayed, which is the reason for the showActions prop.

packages/dfw/src/dfwcomponents/WorkOrderCell/PropertyList.tsx
```diff
@@ -0,0 +1,95 @@
...
+    valueColor,
+    currentDate,
+    overwriteLabel = {},
+}: PropertyListProps) => {
+    const requiredEnd = workOrder.requiredEnd ? new Date(workOrder.requiredEnd) : null;
+
+    const formatDate = (dateString: string) => moment(dateString).format("DD.MM.YYYY");
+    const token = useToken();
+
+    const getLabel = (key: keyof WorkOrder, combinedDates: string | null) => {
```
tormodAase, line 39: `combinedDates` isn't necessary to have as an argument. `getLabel` has access to it

packages/dfw/src/dfwcomponents/WorkOrderCell/StatusIcon.tsx
```diff
@@ -0,0 +1,49 @@
...
+    icon: IconName;
+    label: string;
+    textColor: Color;
+    iconColor: Color;
+};
+
+type StatusIconProps = {
+    statusConfig: StatusConfig;
+    hideLabel?: boolean;
+};
```
tormodAase, line 22: This can probably be simplified by removing the `hideLabel` prop, and making `label` optional. If `label` is undefined, don't display it.

You can probably also simplify this even further by using `StatusConfig` as props directly.


```suggestion
type StatusIconProps = {
    icon: IconName;
    label?: string;
    textColor: Color;
    iconColor: Color;
};
```

packages/dfw/src/dfwcomponents/WorkOrderCell/types.ts
```diff
@@ -0,0 +1,40 @@
...
+    requiredEnd?: string;
+    workCenterId?: string;
+    functionalLocation?: string;
+};
+
+export type WorkOrderCellProps = {
+    showSymbols?: boolean | "column" | "row";
```
tormodAase, line 19: This prop can be hard to understand. I would split it into two props.
```suggestion
    showSymbols?: boolean;
    iconDirection?: "column" | "row"
```

You can make `"column"` the default value of iconDirection

packages/dfw/src/dfwcomponents/WorkOrderCell/types.ts
```diff
@@ -0,0 +1,40 @@
...
+    onCompleteButtonPress?: () => void;
+    onTecoButtonPress?: () => void;
+} & WorkOrder;
+
+export type StatusConfig = {
+    icon: IconName;
+    label: string;
+    textColor: Color;
+    iconColor: Color;
+};
```
tormodAase, line 40: This can probably be removed? `StatusIcon.tsx` already has the same type inside of it

tormodAase (changes requested):
Looks good! I left some comments you can take a look at! Also, would highly encourage you to have someone else with business logic expertise take a look at it as well!

packages/dfw/src/dfwcomponents/WorkOrderCell/WorkOrderCell.tsx
tormodAase: Yes, but the buttons' visibility are also dependent on `onStartButtonPress`, `onCompleteButtonPress`, and `onTecoButtonPress` respectively. To me that seems a bit redundant.

If a developer sets `showActions.startButton` to true, but `onStartButtonPress` is not defined yet, then the button will not be displayed. I would be confused if that happened to me

Jonas-Omdal: ahh. I see

tormodAase (approved):

tormodAase (approved):

Issue #625, opened 2024-09-12
Spacing issue in the WorkOrderCell component


Jonas-Omdal:
### Package / App

MAD DFW

### Version

0.4.0

### What happened?

![image](https://github.com/user-attachments/assets/51e99be0-afee-4265-afb3-0b96c2b813c5)
- Too much space here
- The `WorkOrderType` should have a fallback value for when the work order type do not exist.

PR #626, opened 2024-09-12
fix(workordercell): remove excess padding and add work order type

Jonas-Omdal:
**Key changes:**
- Added new value to the `WorkOrderType`, "unkown". 
- Fix padding issue for when the `maintenanceType` is not being used.

Fixes: #625

### DIFF
--- a/packages/dfw/src/dfwcomponents/WorkOrderCell/types.ts
+++ b/packages/dfw/src/dfwcomponents/WorkOrderCell/types.ts
@@ -10,7 +10,8 @@ export type WorkOrderType =
     | "PM06"
     | "PM10"
     | "PM15"
-    | "PM20";
+    | "PM20"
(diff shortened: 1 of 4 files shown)


packages/dfw/src/dfwcomponents/WorkOrderCell/types.ts
```diff
@@ -10,7 +10,8 @@ export type WorkOrderType =
     | "PM06"
     | "PM10"
     | "PM15"
-    | "PM20";
+    | "PM20"
+    | "Unknown";
```
HaakonSvane, line 14: nitpicky, but maybe this should be lowercased?

HaakonSvane (dismissed):
👍

HaakonSvane (approved):

Issue #631, opened 2024-09-23
Make the WorkOrderCell.Navigation component swipeable


Jonas-Omdal:
### Package / App

MAD DFW

### Description

The WorkOrderCell.Navigation needs to be swipeable so that we can add and remove bookmarks in the work orders app.

PR #632, opened 2024-09-23
feat(workordercell): make the navigation variant swipeable

Jonas-Omdal:
**Key changes:**
- Make the `WorkOrderCell.Navigation` swipeable.
- Changed label `Functional Location` to `Functional location` and adjusted its position in the property list. According to Figma, it should be placed further down.
https://www.figma.com/design/S4NQWezdYGuVz5aAp4xAo4Bl/WorkOrders?node-id=29970-74977&node-type=canvas&t=7zBMgCSo3bSXpvLu-0 
- Added new value, `Operations from filter`,  to the `WorkOrderPropertyList`.

Fixes: #631

### DIFF
--- a/.changeset/afraid-peaches-jam.md
+++ b/.changeset/afraid-peaches-jam.md
@@ -1,0 +1,5 @@
+---
+"@equinor/mad-dfw": patch
+---
+
+The `WorkOrderCell.Navigation` component is now swipeable.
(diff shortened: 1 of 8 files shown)


HaakonSvane (changes requested):
Looks good 👍 
Just the stuff we discussed on Slack regarding the operations from filters prop

PR #634, opened 2024-09-25
chore: Update wrapper to 1.33.0

fnKode:
* Removed trailing slash from the url in `request.ts`:
from
```const url = `${config.BASE}${path}` + (path.endsWith('/') ? '' : '/');```
to
```const url = `${config.BASE}${path}`;```
This fixes an issue where requests to the mAPI would fail due to the trailing slash.
* Updated the wrapper to use mAPI version 1.33.0


HaakonSvane (approved):
Looks fresh 🔥

PR #641, opened 2024-10-07
fix(workordercell): fix validation for the not started icon

Jonas-Omdal:
**Key changes:**
- The validation for showing the start job button was incorrect and should not be shown as intended.
- Made the text in the cell selectable

### DIFF
--- a/packages/dfw/src/dfwcomponents/WorkOrderCell/utils.ts
+++ b/packages/dfw/src/dfwcomponents/WorkOrderCell/utils.ts
@@ -64,7 +64,7 @@ export const getStatusIconsAndLabels = (
         iconsAndLabels.push(getStatusIconConfig("productionCritical")!);
     }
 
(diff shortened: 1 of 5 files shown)


packages/dfw/src/dfwcomponents/WorkOrderCell/utils.ts
```diff
@@ -26,58 +49,29 @@ export const getStatusIconsAndLabels = (
...
-        !activeStatuses?.includes("RDOP") &&
-        !activeStatuses?.includes("TECO")
-    ) {
-        iconsAndLabels.push({
-            icon: "circle-outline",
-            label: "Not started",
-            textColor: "textTertiary",
-            iconColor: "textPrimary",
-        });
+    if (activeStatuses.includes("RDOP") && activeStatuses.includes("TECO")) {
```
fnKode: I think `||` should be used here

fnKode (approved):

Issue #648, opened 2024-10-09
The text values in the WorkOrderCell should wrap


Jonas-Omdal:
### Package / App

MAD Components

### Version

0.5.4

### What happened?

The values in the property rows used in the WorkOrderCell and WorkOrderCell.Navigation should have a prop that allows them to wrap.

PR #649, opened 2024-10-09
fix(workordercell): make text values wrap

Jonas-Omdal:
New prop, `wrapValues` that make text values wrap in the WorkOrderCell and WorkOrderCell.Navigation.
Aslo found and removed a {" "} in the WorkOrderCell.Navigation that made the app crash on IoS.

Fixes: #648


HaakonSvane (approved):
👍

Issue #653, opened 2024-10-15
Add bookmark icons to the navigation workorder cell


Jonas-Omdal:
### Package / App

MAD DFW

### Description

The WorkOrderCell.Navigation is currently missing the bookmark icon. 
![Image](https://github.com/user-attachments/assets/e51947a2-036a-4a0b-8ef3-de86759a10da)

PR #654, opened 2024-10-15
feat(workordercell): add bookmark icons to the navigation variant

Jonas-Omdal:
**Key changes:**
- Added a new bookmarked prop to the WorkOrderCell.Navigation component, which can be used to indicate whether the work order is bookmarked or not. Additionally, new icons have been included to reflect the bookmarked state.

Fixes: #653

### DIFF
--- a/packages/dfw/src/dfwcomponents/WorkOrderCell/WorkOrderCellNavigation.tsx
+++ b/packages/dfw/src/dfwcomponents/WorkOrderCell/WorkOrderCellNavigation.tsx
@@ -19,7 +19,7 @@ type WorkOrderCellNavigationProps = WorkOrderCellProps & {
     onPress: () => void;
     leftSwipeGroup?: SwipeGroup;
     rightSwipeGroup?: SwipeGroup;
-    bookmarked?: boolean;
+    isBookmarked?: boolean;
 };
 
 export const WorkOrderCellNavigation = ({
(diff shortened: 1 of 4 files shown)


packages/dfw/src/dfwcomponents/WorkOrderCell/WorkOrderCellNavigation.tsx
```diff
@@ -19,6 +19,7 @@ type WorkOrderCellNavigationProps = WorkOrderCellProps & {
     onPress: () => void;
     leftSwipeGroup?: SwipeGroup;
     rightSwipeGroup?: SwipeGroup;
+    bookmarked?: boolean;
```
HaakonSvane, line 22: What do you think about naming this `isBookmarked` instead? Very minor, but I think (?) other boolean props we use are named in the same way.

HaakonSvane (dismissed):
👍

packages/dfw/src/dfwcomponents/WorkOrderCell/WorkOrderCellNavigation.tsx
Jonas-Omdal: Ye, i agree. i renamed stuff to isBookmarked instead now.

HaakonSvane (approved):

PR #657, opened 2024-10-16
chore: set max 4 lines of text in toasts

fnKode:
In some cases we are unable to display the entire toast message using two lines. This PR increases the maximum to 4 lines.


tormodAase (approved):
If it looks fine then go for it!

Issue #659, opened 2024-10-17
WorkOrderCell Navigation has issues on smaller devices


Jonas-Omdal:
### Package / App

MAD DFW

### Version

0.5.6

### What happened?

![Image](https://github.com/user-attachments/assets/a2d92a5b-6f3e-49c0-822d-569754dacd45)

The icons should wrap to a new line

PR #660, opened 2024-10-17
fix(workordercell): make icons wrap on smaller devices

Jonas-Omdal:
Icons now shows on a new line on smaller devices in the WorkOrderCell.Navigation component. 

Note, this only happens on smaller devices, such as phones. On tablets the show as before.

Fixes: #659


fnKode (approved):
🚀

Issue #477, opened 2024-04-09
Add `busy` support to `Button.Icon`


HaakonSvane:
### Package / App

MAD Components

### Description

The normal button has a `busy` state that replaces the button content with a dot progress loading indicator.
This is not supported for `Button.Icon`.

PR #670, opened 2024-10-23
feat(button): add loading indicator in the icon button

Jonas-Omdal:
Added new prop, `busy`, in the `Button.Icon` component. When true it shows a loading indicator.

Fixes: #477

### DIFF
--- a/packages/components/src/components/Button/IconButton.tsx
+++ b/packages/components/src/components/Button/IconButton.tsx
@@ -61,24 +61,22 @@ export const IconButton = forwardRef<View, IconButtonProps & ViewProps>(
         });
 
         return (
-            <View>
-                <View ref={ref} style={[styles.colorContainer, rest.style]}>
-                    <PressableHighlight
-                        id={rest.id}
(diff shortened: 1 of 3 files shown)


packages/components/src/components/Button/IconButton.tsx
```diff
@@ -5,6 +5,7 @@ import { EDSStyleSheet } from "../../styling";
 import { getBackgroundColorForButton } from "../../utils/getBackgroundColorForButton";
 import { Icon, IconName } from "../Icon";
 import { PressableHighlight } from "../PressableHighlight";
+import { DotProgress } from "../ProgressIndicator";
```
HaakonSvane, line 8: In EDS, busy icon buttons use the CircularProgress component to indicate its state instead of dots. Maybe change to using this? 

Screenshot 2024-10-24 at 12 09 19

HaakonSvane (changes requested):
👍 
See my comment on the type of indicator

HaakonSvane (approved):
👍

Issue #674, opened 2024-11-01
onPress in cell swipeGroups causes errors


Jonas-Omdal:
### Package / App

MAD Components

### Version

0.15.1

### What happened?

Every time you trigger onPress in a cell swipeGroup, this happens:

ERROR  Swipeable method unavailable. Make sure you are running this method inside a `SwipeableWithContext` component. Otherwise, contact maintainers of `@equinor/mad-components`

Example:
`
           leftSwipeGroup={[
                        {
                            title: "example",
                            color: "primary",
                            onPress: methods => methods.openRight(),
                        },
                    ]}
`

PR #675, opened 2024-11-01
fix(cell): fix swipe group press error

Jonas-Omdal:
Fixes: #674


HaakonSvane (approved):
Tested on web and iOS. Seems to be working well 👌

Issue #164, opened 2023-08-17
`Button`: `fullWidth` prop


dansvend:
### Package / App

MAD Components

### Description

Add a `fullWidth` prop to `Button` like EDS has for their [Button](https://storybook.eds.equinor.com/?path=/docs/inputs-button-button--docs).

PR #678, opened 2024-11-06
feat(Button): add new prop fullwidth

Jonas-Omdal:
Added fullwidth prop to the button component 
Fixes: #164


HaakonSvane:
Two cents:
Do we need this functionality or can it be achieved in another way? I was hoping that the current style prop on button allowed some way of achieving the same results?

HaakonSvane (approved):

PR #690, opened 2024-11-27
docs: fix broken link in navigation readme

fnKode:


.changeset/cool-spies-return.md
```diff
@@ -0,0 +1,5 @@
+---
+"@equinor/mad-navigation": patch
+---
+
+Fix broken link
```
tormodAase, line 5: Probably don't need a changeset for this one

tormodAase: Or maybe the npm registry wont update the description unless we version bump? @HaakonSvane

tormodAase (commented):
Not sure if we need a changeset. Maybe @HaakonSvane can give some input on whether we need a changeset or not

.changeset/cool-spies-return.md
HaakonSvane: Yeah it probably won't see this package as publishable unless we do I think

HaakonSvane (approved):

PR #691, opened 2024-11-27
fix(core): remove-button on service message banner now works again

tormodAase:
Also, we no longer export `CellSwipeItem` from `mad-components`


HaakonSvane (approved):
LGTM 🚀

PR #703, opened 2024-12-17
fix: add disabled prop to x icon

Sandtory:
Add disabled prop to x icon so that it will not let user clear the search field when used

### DIFF
--- a/packages/components/src/components/Search/Search.tsx
+++ b/packages/components/src/components/Search/Search.tsx
@@ -145,7 +145,7 @@ export const Search = ({
                         </View>
                     }
                     rightAdornments={
-                        text ? (
+                        text && disabled ? (
                             <View style={styles.adornmentIconContainer}>
                                 <Button.Icon
                                     id="search-clear-text-button"
@@ -153,7 +153,6 @@ export const Search = ({
                                     color="primary"
                                     variant="ghost"
                                     onPress={handleClearText}
-                                    disabled={disabled}
                                 />
                             </View>
                         ) : null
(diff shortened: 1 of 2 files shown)


packages/components/src/components/Search/Search.tsx
```diff
@@ -153,6 +153,7 @@ export const Search = ({
                                     color="primary"
                                     variant="ghost"
                                     onPress={handleClearText}
+                                    disabled={disabled}
```
HaakonSvane, line 156: Should we maybe instead just don't show this clear button if the search bar is disabled? 
Could potentially just add the disabled check next to the condition for rendering this on line 148 in that case. Thoughts?

HaakonSvane (approved):
Nice 😠 
Check my comment regarding the rendering of the button and change if you think it is good idea.

packages/components/src/components/Search/Search.tsx
Sandtory: yes that is smoother and more highly advanced senior code, i agree wholeheartedly 😡

packages/components/src/components/Search/Search.tsx
```diff
@@ -145,7 +145,7 @@ export const Search = ({
                         </View>
                     }
                     rightAdornments={
-                        text ? (
```
HaakonSvane, line 148: Isn't it the other way around? 
`text && !disabled` ?
Now it seems to only render the clear button if the search contains text AND is disabled

HaakonSvane (changes requested):

HaakonSvane (approved):
😠 perfect

PR #712, opened 2025-01-28
fix: make dialog responsive

### LINKED ISSUE #711: Dialog does not show all content when using large text size
### Package / App

MAD Components

### Version

0.19.2

### What happened?

When adjusting text size on an iPhone in  `Settings` -> `Accessibility` -> `Display & Text Size` -> `Larger Text` -> Enable `Larger Accessibility Sizes` -> Adjust text size to the largest text size -> Open up Chronicles or an app that is using the `Dialog` component or `alert` to trigger the `Dialog`.

The title is clipped.  The user is not able to view all the content since the Dialog does not have an ability to scroll or navigate to next page. The text of the buttons are also clipped.

A user should be able to view the different parts of the dialog, press all the buttons and no text should be clipped.

### Other info

Standard text size on an iPhone 16 Pro Max:
Image

Largest text size on an iPhone 16 Pro Max with `Larger Accessibility Sizes` enabled:
Image

### Relevant log output

```shell

```

dansvend:
Components
- Make dialog title responsive
- Make dialog custom content scrollable
- Make action buttons wrap when necessary
- Avoid clipping of button content in dialog action buttons

Chronicles
- Wrap buttons in button screen
- Make gap and padding in button row responsive

Closes #711

### DIFF
--- a/apps/chronicles/screens/components/components/ButtonScreen.tsx
+++ b/apps/chronicles/screens/components/components/ButtonScreen.tsx
@@ -102,6 +102,8 @@ const themeStyles = EDSStyleSheet.create(theme => ({
     buttonRow: {
         paddingVertical: 8,
         flexDirection: "row",
+        flexWrap: "wrap",
         justifyContent: "space-around",
+        gap: 12,
     },
 }));
--- a/apps/chronicles/screens/components/components/ButtonScreen.tsx
+++ b/apps/chronicles/screens/components/components/ButtonScreen.tsx
@@ -100,10 +100,10 @@ const themeStyles = EDSStyleSheet.create(theme => ({
         paddingVertical: theme.spacing.container.paddingVertical,
     },
     buttonRow: {
-        paddingVertical: 8,
+        paddingVertical: theme.spacing.container.paddingVertical,
         flexDirection: "row",
         flexWrap: "wrap",
         justifyContent: "space-around",
-        gap: 12,
+        gap: theme.spacing.spacer.small,
     },
 }));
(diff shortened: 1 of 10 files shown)


apps/chronicles/screens/components/components/ButtonScreen.tsx
```diff
@@ -102,6 +102,8 @@ const themeStyles = EDSStyleSheet.create(theme => ({
     buttonRow: {
         paddingVertical: 8,
         flexDirection: "row",
+        flexWrap: "wrap",
         justifyContent: "space-around",
+        gap: 12,
```
HaakonSvane, line 107: Is there a token spacing value we can use instead of this?
I see the padding vertical is also a hard-coded number, but I won't bother you with previously written code

HaakonSvane (dismissed):
Looks good. Besides the hard-coded styling value, everything looks lit 🔥

HaakonSvane:
Quick update. Tested on web by resizing browser window and scale and it seems the dialog extends the browser boundaries and is not scrollable. See video


https://github.com/user-attachments/assets/0c30e8a6-3feb-43eb-9348-d9908cd2e47d

apps/chronicles/screens/components/components/ButtonScreen.tsx
dansvend: `gap` and `paddingVertical` now uses responsive values from `theme` in [d069be9](https://github.com/equinor/mad/pull/712/commits/d069be9eaf7473f379c8d6863f189a801bba55ba).

dansvend:
Previous issues should now be fixed 🤞

dansvend:
@HaakonSvane Previous issues should now be fixed 🤞

PR #722, opened 2025-04-02
chore: Update API wrapper to support api version 1.37.0

SUBBH-Equinor:


CaarnHeir (approved):
Looks good/inline with our conversation.

PR #724, opened 2025-04-15
chore: Set Maintenance API wrapper version to 1.37.0

SUBBH-Equinor:


CaarnHeir (approved):
Approved based on teams messages

PR #736, opened 2025-06-23
fix: Remove duplicate links from about section

eqrks:
Removes duplicate links from about section

image
image


Murugan-sudo (approved):
Code is written for removing the repeated link from the page.Code fix the issue.

PR #768, opened 2025-10-30
fix: patches added to allow peer dependencies to upgrade to the lates…

SUBBH-Equinor:
…t patch versions


OGUPEQUINOR (approved):
approved

PR #821, opened 2026-04-21
fix: getAccount function fixed for web

### LINKED ISSUE #820: GetAccount for Web returns null even when the user is logged in
### Package / App

MAD Core

### Version

0.5.6

### What happened?

GetAccount should return valid account data for web when user is logged in

### Other info

_No response_

### Relevant log output

```shell

```

SUBBH-Equinor:


OGUPEQUINOR (approved):
approved

PR #825, opened 2026-05-13
perf: expo and react native upgraded

### LINKED ISSUE #819: Upgrade to latest Expo version (55)
The platform is still running on Expo SDK 53, while the latest framework version is 55. 56 already has a canary release, so it is expected to land rather soon.

The problems we are experiencing running on outdated versions are:
- __Security vulnerabilities__: new vulnerabilities related to our Expo version keeps popping up. Unless Expo ships patches to their old 53 version, we need to manually override sub-dependencies to avoid this problem
- __Can not build on Xcode 26.4__: React native recently encountered issues with building from source due to a change in Xcode. This problem will be resolved in the latest versions of RN + Expo, but does not seem to be planned for older Expo versions. This means we either have to manually downgrade Xcode or temporarily patch the podfile to be able to build our projects. 
- __Bad DevEx__: Outdated versions means we need to refer to outdated documentation for libraries we use. This makes debugging harder for us.

SUBBH-Equinor:


packages/navigation/package.json
SUBBH-Equinor, line 1: change react native to 0.83.6

OGUPEQUINOR (approved):
approved

PR #867, opened 2026-08-21
fix: type support added back to navigation

### LINKED ISSUE #856: Navigation factories are typeless
### Package / App

MAD Core

### Version

latest

### What happened?

After some recent changes in the repository, it seems that generics and typing for the navigation factories (e.g. `createStackNavigatorFactory`) are lost. This means we can not type-hint our navigators with our navigation hierarchy, so TypeScript support is effectively lost.

Here are the changes in the built code for the stack navigation factory:

### Before change

Image

### After change

Image

### Other info

_No response_

### Relevant log output

```shell

```

SUBBH-Equinor:


OGUPEQUINOR (approved):
approved

PR #878, opened 2026-09-10
fix: 854 menu popover rendered below native modals

### LINKED ISSUE #854: Menu popover rendered below native modals
### Package / App

MAD Components

### Version

latest

### What happened?

Quick investigation, it seems that components that rely on the `RootModal` component or any component that portals components to the root DOM nodes render below the native modals. This means menus and other similar components (autocomplete for example) do not work for modals. This bug only seems to affect native navigation containers. See video for example

### Other info

Video showcasing first a working menu rendering in a normal stack hierarchy, then the same menu opened from a modal screen:
https://github.com/user-attachments/assets/1f35affe-76e1-479d-ad94-a8b3dd04cb9c

### Relevant log output

```shell

```

SUBBH-Equinor:
- Replaced 'Portal' in Root Modal component with 'Native Modal' in order to fix the bug where Root Modal components were being rendered below Native Modal components when using Native Navigator

### DIFF
--- a/.changeset/rich-lions-drop.md
+++ b/.changeset/rich-lions-drop.md
@@ -1,0 +1,15 @@
+---
+"@equinor/react-native-mad-tag-ocr": patch
+"@equinor/mad-components": patch
+"@equinor/mad-navigation": patch
+"@equinor/react-native-skia-draw": patch
+"@equinor/mad-insights": patch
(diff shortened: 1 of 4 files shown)


OGUPEQUINOR (approved):
approved

